PR #17, opened 2017-10-17
Fix #16

### LINKED ISSUE #16: Issue installing under Windows
Hi! I've been having a bit of trouble installing `cv-rs` under Windows. I have followed the OpenCV installation tutorial (using downloaded binaries), and have tested it (and found it working) with Python (2.7). I've added basically all the directories inside of the opencv dir, including the ones in `x64` to `PATH`, yet it still seems like the compiler is missing `opencv2\core.hpp` :/ Any help would be appreciated!

---
[Full error/build log](https://gist.github.com/Restioson/31be55c41b41ea643bb0dbfd59642109)

Restioson:
This adds an include for `%OPENCV_DIR%\include` (if it exists) to the GCC build, which should fix #16, so long as the variable is set correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,12 @@ git = "https://github.com/nebgnahz/cv-rs"
 features = [ "gpu" ]
 ```
 
+### Windows
+
+Depending on your install, you might have to set the `%OPENCV_DIR%`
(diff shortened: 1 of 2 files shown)


nebgnahz:
Merged, thanks for the PR!

PR #19, opened 2017-10-17
configure travis

nebgnahz:


Restioson:
Seems to work. Can be merged? Ps: does this fix the can't find opencv_core.lib (etc) bug?

nebgnahz:
Ah, this has been merged (I directly patched the master branch).

I used to use travis-ci.com so that I can debug the project. But this is a public repository, so I have to use travis-ci.org: the only way to debug is trial and error. Therefore it has 11 commits...

Restioson:
Ouch, unfortunately that's quite relatable though. So, should this be fixing the runtime bug under Windows?

Restioson:
Huh. When you were referring to a bug on Windows, I thought you were referring to `can't find opencv_core.lib` (what I get). I've installed it by downloading the zip file, maybe I will reclone the repo and try again, and if that fails I will use choco

nebgnahz:
I see.  So according to [this SO question](https://stackoverflow.com/questions/34867289/opencv-3-x-only-contains-one-lib-opencv-world-lib), binary installers have a single `opencv_world.lib` file. When I was testing on AppVeyor, I saw `opencv_world320.lib`, so presumably choco installed OpenCV 3.2. In this case, `can't find opencv_core.lib` makes sense, because there is no longer `opencv_core.lib`. But I think if one builds OpenCV from source on Windows, there will be separate libraries including `opencv_core.lib`; and this project will fail in that case.

PR #27, opened 2018-01-02
autodetecting of the opencv version (Windows)

Pzixel:


build.rs
```diff
@@ -13,8 +13,23 @@ fn opencv_include() -> String {
...
+                .filter_map(|entry| entry.ok())
+                .find(|entry| {
+                    let file_name = entry.file_name().to_string_lossy().into_owned();
+                    file_name.starts_with("opencv_world") && !file_name.ends_with("d.lib")
+                });
+            if let Some(opencv_world) = opencv_world_entry {
+                let opencv_world = opencv_world.file_name();
+                let opencv_world = std::path::Path::new(&opencv_world);
+                println!("cargo:rustc-link-search=native={}", dir);
+                println!("cargo:rustc-link-lib={}", opencv_world.file_stem().unwrap().to_string_lossy());
```
Pzixel: unwrap is OK here because file_stem returns None if file_name is None which is impossible in this case.

Pzixel: Sorry, it's 5.30AM in my timezone so I'm goint to take some sleep. Can reply after 10 hours or something.

build.rs
```diff
@@ -13,8 +13,23 @@ fn opencv_include() -> String {
...
-        println!("cargo:rustc-link-lib=opencv_world330");
+        if let Ok(mut files) = std::fs::read_dir(&dir) {
+            let opencv_world_entry = files
+                .filter_map(|entry| entry.ok())
+                .find(|entry| {
+                    let file_name = entry.file_name().to_string_lossy().into_owned();
+                    file_name.starts_with("opencv_world") && !file_name.ends_with("d.lib")
+                });
+            if let Some(opencv_world) = opencv_world_entry {
+                let opencv_world = opencv_world.file_name();
```
nebgnahz: Is it possible to use [DirEntry::path](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.path) without instead of converting to `OsString` and then back to `Path`?

nebgnahz:
While we are at it, can you also update README, [Windows](https://github.com/nebgnahz/cv-rs#windows) section?

Pzixel:
Of course, I was thinking about it myself.

nebgnahz:
Merged! Thanks!

PR #30, opened 2018-01-05
MinGW support

Pzixel:


build.rs
```diff
@@ -18,12 +18,14 @@ fn opencv_link() {
...
                 .find(|entry| {
                     let file_name = entry.file_name().to_string_lossy().into_owned();
-                    file_name.starts_with("opencv_world") && !file_name.ends_with("d.lib")
+                    (file_name.starts_with("opencv_world") || file_name.starts_with("libopencv_world")) && !file_name.ends_with("d.lib")
                 });
             if let Some(opencv_world) = opencv_world_entry {
-                let opencv_world = opencv_world.path();
+                let opencv_world = opencv_world.file_name();
+                let opencv_world = opencv_world.into_string().unwrap();
+                let opencv_without_extension = opencv_world.trim_right_matches(|c: char| !c.is_numeric());
```
Pzixel: `file_stem` is not working properly with "double-extension" `.dll.a`. Because we know the shape of the exe `(lib)opencw_world<some_numbers>.extensiontrrash` we just trim everything after version number.

build.rs
```diff
@@ -18,12 +18,14 @@ fn opencv_link() {
...
                 .find(|entry| {
                     let file_name = entry.file_name().to_string_lossy().into_owned();
-                    file_name.starts_with("opencv_world") && !file_name.ends_with("d.lib")
+                    (file_name.starts_with("opencv_world") || file_name.starts_with("libopencv_world")) && !file_name.ends_with("d.lib")
                 });
             if let Some(opencv_world) = opencv_world_entry {
-                let opencv_world = opencv_world.path();
+                let opencv_world = opencv_world.file_name();
+                let opencv_world = opencv_world.into_string().unwrap();
+                let opencv_world_without_extension = opencv_world.trim_right_matches(|c: char| !c.is_numeric());
```
Pzixel: `file_stem` is not working properly with "double-extensions" such as `.dll.a`. Because we know the shape of the exe `(lib?)opencw_world<some_numbers>.extension` we just trim everything after version number. We are also free to call `unwrap` after `into_string` because this shape cannot contain invalid UTF8 data.

build.rs
```diff
@@ -18,12 +18,14 @@ fn opencv_link() {
...
                 println!("cargo:rustc-link-search=native={}", dir);
-                println!("cargo:rustc-link-lib={}", opencv_world.file_stem().unwrap().to_string_lossy());
+                println!("cargo:rustc-link-lib={}", opencv_world_without_extension);
```
Pzixel: `file_stem` is not working properly with "double-extensions" such as `.dll.a`. Because we know the shape of the exe `(lib?)opencw_world<some_numbers>.extension` we just trim everything after version number. We are also free to call `unwrap` after `into_string` because this shape cannot contain invalid UTF8 data.

build.rs
```diff
@@ -18,12 +18,14 @@ fn opencv_link() {
                 .filter_map(|entry| entry.ok())
                 .find(|entry| {
                     let file_name = entry.file_name().to_string_lossy().into_owned();
-                    file_name.starts_with("opencv_world") && !file_name.ends_with("d.lib")
+                    (file_name.starts_with("opencv_world") || file_name.starts_with("libopencv_world")) && !file_name.ends_with("d.lib")
                 });
             if let Some(opencv_world) = opencv_world_entry {
-                let opencv_world = opencv_world.path();
+                let opencv_world = opencv_world.file_name();
```
nebgnahz: This seems complicated enough to justify some comments as clarifications in the code. Maybe just explain it is possible to have `.dll.a` or `.lib`, so we are trim everything until we hit the number in 
version number.

Pzixel: I think `opencv_world_without_extension` is kinda self-explaining name. I can add a comment here, but I'm not sure if we need it anyway.

nebgnahz: Thanks for the tip. I thought my comments are somewhat optional to address so I approved the PR. For future PRs, I will **not** approve unless I will merge immediately.

nebgnahz:
Merged. Thanks!

PR #31, opened 2018-01-06
MSER

Pzixel:
It also contains several improvements in logging for Windows. I had to write it for myself, so it may be useful for useres anyway.

### DIFF
--- a/native/opencv-gpu.cc
+++ b/native/opencv-gpu.cc
@@ -42,7 +42,7 @@ using CV_GPU_HOG = cv::Ptr<cv::cuda::HOG>;
 
 GpuHog* cv_gpu_hog_default() {
     auto hog = cv::cuda::HOG::create();
-    return reinterpret_cast<GpuHog*>(new CV_GPU_HOG(cv::cuda::HOG::create()));
+    return reinterpret_cast<GpuHog*>(new CV_GPU_HOG(hog));
 }
 
 GpuHog* cv_gpu_hog_new(Size2i win_size, Size2i block_size,
--- a/native/opencv-wrapper.cc
+++ b/native/opencv-wrapper.cc
@@ -605,14 +605,14 @@ CMSER* cv_mser_new(int delta,
                    double min_margin,
                    int edge_blur_size) {
     cv::Ptr<cv::MSER> result = cv::MSER::create(delta,
-                                                 min_area,
-                                                 max_area,
-                                                 max_variation,
-                                                 min_diversity,
-                                                 max_evolution,
-                                                 area_threshold,
-                                                 min_margin,
-                                                 edge_blur_size);
+                                                min_area,
+                                                max_area,
+                                                max_variation,
+                                                min_diversity,
+                                                max_evolution,
+                                                area_threshold,
+                                                min_margin,
+                                                edge_blur_size);
     return reinterpret_cast<CMSER*>(new cv::Ptr<cv::MSER>(result));
 }
 
@@ -622,7 +622,7 @@ void cv_mser_drop(CMSER* cmser) {
     mser = nullptr;
 }
 
-void cv_mser_detectRegions(CMSER* cmser, CvMatrix* image, VecPoints* msers, VecRect* bboxes) {
+void cv_mser_detect_regions(CMSER* cmser, CvMatrix* image, VecPoints* msers, VecRect* bboxes) {
     cv::Ptr<cv::MSER>* mser = reinterpret_cast<cv::Ptr<cv::MSER>*>(cmser);
     cv::Mat* mat = reinterpret_cast<cv::Mat*>(image);
     std::vector<std::vector<cv::Point>> msers_vector;
--- a/native/opencv-wrapper.cc
+++ b/native/opencv-wrapper.cc
@@ -6,6 +6,7 @@
 #include <opencv2/imgproc.hpp>
 #include <opencv2/objdetect.hpp>
 #include <opencv2/video/tracking.hpp>
+#include <opencv2/features2d.hpp>
 
 EXTERN_C_BEGIN
 
@@ -113,6 +114,22 @@ void cv_vec_of_rect_drop(VecRect* v) {
         free(v->array);
         v->array = nullptr;
         v->size = 0;
+    }
+}
+
+void cv_vec_of_point_drop(VecPoint* pv) {
+    if (pv->array != nullptr) {
+        free(pv->array);
+        pv->array = nullptr;
+        pv->size = 0;
+    }
+}
+
+void cv_vec_of_points_drop(VecPoints* pvs) {
+    if (pvs->array != nullptr) {
+        free(pvs->array);
+        pvs->array = nullptr;
+        pvs->size = 0;
     }
 }
 
@@ -572,6 +589,49 @@ RotatedRect cv_camshift(CvMatrix* c_bp_image, Rect crect,
     c_rr.size.height = rr.size.height;
     c_rr.angle = rr.angle;
     return c_rr;
+}
+
+// =============================================================================
+//   MSER
+// =============================================================================
+
+CMSER* cv_mser_new(int delta,
+                   int min_area,
+                   int max_area,
+                   double max_variation,
+                   double min_diversity,
+                   int max_evolution,
+                   double area_threshold,
+                   double min_margin,
+                   int edge_blur_size) {
+    cv::Ptr<cv::MSER> result = cv::MSER::create(delta,
+                                                 min_area,
+                                                 max_area,
+                                                 max_variation,
+                                                 min_diversity,
+                                                 max_evolution,
+                                                 area_threshold,
+                                                 min_margin,
+                                                 edge_blur_size);
+    return reinterpret_cast<CMSER*>(new cv::Ptr<cv::MSER>(result));
+}
+
+void cv_mser_drop(CMSER* cmser) {
+    cv::Ptr<cv::MSER>* mser = reinterpret_cast<cv::Ptr<cv::MSER>*>(cmser);
+    delete mser;
+    mser = nullptr;
+}
+
+void cv_mser_detectRegions(CMSER* cmser, CvMatrix* image, VecPoints* msers, VecRect* bboxes) {
+    cv::Ptr<cv::MSER>* mser = reinterpret_cast<cv::Ptr<cv::MSER>*>(cmser);
+    cv::Mat* mat = reinterpret_cast<cv::Mat*>(image);
+    std::vector<std::vector<cv::Point>> msers_vector;
+    std::vector<cv::Rect> bboxes_vector;
+
+    mser->get()->detectRegions(*mat, msers_vector, bboxes_vector);
+
+    vec_points_cxx_to_c(msers_vector, msers);
+    vec_rect_cxx_to_c(bboxes_vector, bboxes);
 }
 
 EXTERN_C_END
--- a/native/opencv-wrapper.cc
+++ b/native/opencv-wrapper.cc
@@ -127,6 +127,9 @@ void cv_vec_of_point_drop(VecPoint* pv) {
 
 void cv_vec_of_points_drop(VecPoints* pvs) {
     if (pvs->array != nullptr) {
+        for (size_t i = 0; i < pvs->size; ++i) {
+            cv_vec_of_point_drop(&pvs->array[i]);
+        }
         free(pvs->array);
         pvs->array = nullptr;
         pvs->size = 0;
--- a/native/opencv-wrapper.h
+++ b/native/opencv-wrapper.h
@@ -261,7 +261,7 @@ CMSER* cv_mser_new(int delta,
                    double min_margin,
                    int edge_blur_size);
 void cv_mser_drop(CMSER* cmser);
-void cv_mser_detectRegions(CMSER* cmser, CvMatrix* image, VecPoints* msers, VecRect* bboxes);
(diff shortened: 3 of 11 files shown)


native/opencv-gpu.cc
```diff
@@ -42,7 +42,7 @@ using CV_GPU_HOG = cv::Ptr<cv::cuda::HOG>;
 
 GpuHog* cv_gpu_hog_default() {
     auto hog = cv::cuda::HOG::create();
-    return reinterpret_cast<GpuHog*>(new CV_GPU_HOG(cv::cuda::HOG::create()));
+    return reinterpret_cast<GpuHog*>(new CV_GPU_HOG(hog));
```
Pzixel: I have found this double creation of `HOG` and fixed it just in case.

src/mser.rs
```diff
@@ -0,0 +1,171 @@
...
+        MSER::new(
+            self.delta.unwrap_or(5),
+            self.min_area.unwrap_or(60),
+            self.max_area.unwrap_or(14400),
+            self.max_variation.unwrap_or(0.25),
+            self.min_diversity.unwrap_or(0.2),
+            self.max_evolution.unwrap_or(200),
+            self.area_threshold.unwrap_or(1.01),
+            self.min_margin.unwrap_or(0.003),
+            self.edge_blur_size.unwrap_or(5),
```
Pzixel: Default values are taken from here: https://docs.opencv.org/trunk/d3/d28/classcv_1_1MSER.html#a136c6b29fbcdd783a1003bf429fa17ed

Pzixel:
Travis is not able to build this guy:

> /home/travis/build/nebgnahz/cv-rs/target/debug/deps/libcv-218dbf6ab2eb2901.rlib(opencv-wrapper.o): In function `cv_mser_new':
>           /home/travis/build/nebgnahz/cv-rs/native/opencv-wrapper.cc:615: undefined reference to `cv::MSER::create(int, int, int, double, double, int, double, double, int)'
>           collect2: error: ld returned 1 exit status

Can you help with it? It seems that `build.rs` for linux doesn't include all required files. I have tried to add following:
```
println!("cargo:rustc-link-lib=opencv_features2d");
```

but it didn't help.

native/opencv-gpu.cc
nebgnahz: Good catch.

nebgnahz:
Looks like the MSER detection returns different value on different platforms and different versions. I have tested on the following platforms/versions:

Linux, OpenCV 3.4 passes
Linux, OpenCV 3.1 => 299
OS X, OpenCV 3.1 => 133
Linux, OpenCV 3.3 (Travis) => 252

I don't have a lot of knowledge about MSER. Maybe just test the returned vector has more than one element, so that we know MSER is correctly wrapped and can return something?

Pzixel:
Ok, let's check it. Not really resilent, but at least it would work.

native/opencv-wrapper.cc
```diff
@@ -574,4 +591,47 @@ RotatedRect cv_camshift(CvMatrix* c_bp_image, Rect crect,
...
+    return reinterpret_cast<CMSER*>(new cv::Ptr<cv::MSER>(result));
+}
+
+void cv_mser_drop(CMSER* cmser) {
+    cv::Ptr<cv::MSER>* mser = reinterpret_cast<cv::Ptr<cv::MSER>*>(cmser);
+    delete mser;
+    mser = nullptr;
+}
+
+void cv_mser_detectRegions(CMSER* cmser, CvMatrix* image, VecPoints* msers, VecRect* bboxes) {
```
nebgnahz: s/cv_mser_detectRegions/cv_mser_detect_regions?

More consistent with the rest of the naming conventions.

native/opencv-wrapper.h
```diff
@@ -235,6 +247,22 @@ void cv_term_criteria_drop(CTermCriteria* c_criteria);
...
+                   int min_area,
+                   int max_area,
+                   double max_variation,
+                   double min_diversity,
+                   int max_evolution,
+                   double area_threshold,
+                   double min_margin,
+                   int edge_blur_size);
+void cv_mser_drop(CMSER* cmser);
+void cv_mser_detectRegions(CMSER* cmser, CvMatrix* image, VecPoints* msers, VecRect* bboxes);
```
nebgnahz: cv_mser_detect_regions? See above

native/utils.cc
```diff
@@ -19,3 +19,24 @@ void vec_double_cxx_to_c(const std::vector<double>& cxx_vec_double,
     vd->array = (double*) malloc(num * sizeof(double));
     ::memcpy(vd->array, cxx_vec_double.data(), num * sizeof(double));
 }
+
+void vec_point_cxx_to_c(const std::vector<cv::Point>& cxx_vec_point,
+                         VecPoint* vp) {
```
nebgnahz: Remove the extra space for alignment.

src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub mod videoio;
 pub mod highgui;
 pub mod video;
 pub mod objdetect;
+pub mod mser;
```
nebgnahz: mod features2d? So that this is consistent with OpenCV.

src/mser.rs
```diff
@@ -0,0 +1,171 @@
...
+                                           max_variation,
+                                           min_diversity,
+                                           max_evolution,
+                                           area_threshold,
+                                           min_margin,
+                                           edge_blur_size) };
+        MSER { value: mser }
+    }
+
+    /// Detect MSER regions.
```
nebgnahz: s/Detect/Detects

src/mser.rs
```diff
@@ -0,0 +1,171 @@
...
+                   max_area: i32,
+                   max_variation: f64,
+                   min_diversity: f64,
+                   max_evolution: i32,
+                   area_threshold: f64,
+                   min_margin: f64,
+                   edge_blur_size: i32
+    ) -> *mut CMSER;
+    fn cv_mser_drop(cmser: *mut CMSER);
+    fn cv_mser_detectRegions(cmser: *mut CMSER, image: *mut CMat, msers: *mut CVecOfPoints, bboxes: *mut CVecOfRect);
```
nebgnahz: Do you need a mutable reference to `image` here?

Pzixel:
After some investigation I have found that MSER don't fit my needs 😃 

You are working with CV so I assume you have some expirience in this field. I'd like to compare two images if they are similar or not. I'd like to check if this image is already exist in my database, but it also should be resilent to affine transformation, simple rotations etc. What OpenCV algorithm fits these needs? I think I can implement it too, but I'm not sure which I must take :)

src/mser.rs
```diff
@@ -0,0 +1,171 @@
...
+            min_diversity: None,
+            max_evolution: None,
+            area_threshold: None,
+            min_margin: None,
+            edge_blur_size: None,
+        }
+    }
+
+    /// Replace current delta with specified value
+    pub fn delta(mut self, value: i32) -> Self{
```
nebgnahz: Space in `Self {`.

The whole project should have been formatted with `cargo fmt`... But that should happen in a different PR.

native/opencv-wrapper.cc
```diff
@@ -116,6 +117,22 @@ void cv_vec_of_rect_drop(VecRect* v) {
...
+    if (pv->array != nullptr) {
+        free(pv->array);
+        pv->array = nullptr;
+        pv->size = 0;
+    }
+}
+
+void cv_vec_of_points_drop(VecPoints* pvs) {
+    if (pvs->array != nullptr) {
+        free(pvs->array);
```
nebgnahz: Here, each array element may contain a pointer (as in `VecPoint`). Wouldn't `free` here cause memory leak?

nebgnahz:
My friend Kaifei has used SURF for his project [SnapLink](https://people.eecs.berkeley.edu/~kaifei/data/ubicomp_18.pdf), which matches a new photo to an existing database: sounds like what you want to do.

native/opencv-wrapper.cc
Pzixel: No, `VecPoint` is `Drop` itself. We only must consider our own memory. If you try to free some `VecPoints` here you get SEGFAULT because of double memory free.

src/mser.rs
Pzixel: Documentation says `Detect`: https://de.mathworks.com/help/vision/ref/detectmserfeatures.html?requestedDomain=www.mathworks.com

native/opencv-wrapper.cc
Pzixel: I think we must follow original naming convention which is `detectRegions`. Otherwise it may be hard to find a corresponding method if you know original one.

src/mser.rs
Pzixel: @nebgnahz if you want to use `cargo fmt` I also suggest to create `rustfmt.toml` to persist non-default rules (if any) in the repo itself.

Pzixel:
@nebgnahz thank you, but[ looking on the interface](https://docs.opencv.org/ref/2.4/de/d58/classcv_1_1SURF.html) of SURF I don't see any corresponding methods. I can calculate points of interest, but it actually doesn't compare anything. I wonder how your friend is doing it.

About MR itself: I have commented every discutable comment.

native/opencv-wrapper.cc
nebgnahz: But in Rust, it's not `Vec<VecPoint>`. It's `array: *mut CVecOfPoint`. So the `Drop` for `VecPoint` is not called because Rust doesn't know how to destroy a pointer.

nebgnahz:
Right, SURF just returns you with features. To compare, you will need to calculate the distance between two feature vectors. My friend uses KD-Tree to find the closest nearest neighbor (see Fig. 7).

src/mser.rs
nebgnahz: Sure!

src/mser.rs
nebgnahz: Ok...

native/opencv-wrapper.cc
nebgnahz: Ok, I don't have a strong opinion here.

Personally, I think the conversion from `OriginalMethodName` to `original_method_name` is not a hard one. It's just my preference. You don't have to change.

native/opencv-wrapper.cc
Pzixel: If you allocate `VecPoints` with say 10 values, and you drop it, you can see 10 calls of `VecPoint.drop` and one `VerPoints.drop`. Rust calls `drop` for all `VecPoint` itself and it destroys `VecPoints` only after it. You must not interfer with other types that implement `drop`.

Pzixel:
Ok, I'l keep it in mind. I think I'm going to something simpler like histogram comparision, if it gives good results.

native/opencv-wrapper.cc
Pzixel: If you are in doubt, just create a test method and try to allocate lots of `VecPoints` and check if there is any leak.

nebgnahz: Yes, I've tested: [code here](https://github.com/nebgnahz/cv-rs/commit/876215603dbb25172352f20f849f0888e89d273b).

Running for a few seconds, the tests program has used up to 3-5G memory.

Pzixel: @nebgnahz wow. Ok, need some investigation here. Sounds weird: it should free all the memory,,,

Pzixel: Hmm, it seems that we must dealloc C++ code. Rust `drop` is fine.

nebgnahz: `CVecPoint::drop` is never called (see https://github.com/nebgnahz/cv-rs/commit/b8084b812500d9d98e6ccb43582a61e0212819c3 when I just use `unimplemented{}` there.

Pzixel: @nebgnahz problem is in `vec_rect_cxx_to_c` and its counterparts because they are copying data instead of moving it. You can see the same memory leak for `cv_hog_detect`

nebgnahz: No, the problem is `Vec<Vec<Point>>` has two layers: the first layer is freed correctly, the second layer currently is not. `Vec<Rect>` in HOG is fine since it only has one layer.

The following `cv_vec_of_points_drop` has the right fix.

```
void cv_vec_of_points_drop(VecPoints* pvs) {
    if (pvs->array != nullptr) {
        // Free internal VecPoints
        for (size_t i = 0; i < pvs->size; i++) {
            cv_vec_of_point_drop(&pvs->array[i]);
        }
        free(pvs->array);
        pvs->array = nullptr;
        pvs->size = 0;
    }
}
```

Pzixel: @nebgnahz yep, I had this code but I deleted it for some reason. You convienced me.

However, there is a proble. You see, `std::vector` is never freed. We copy `std::Vector<std::Vector<Point2i>>` to `CVecOfPoints`, but we don't free the source. Your code is showing this problem: fixed `drop` won't save us from this memory leak.

nebgnahz: Yes it is a memory copy (so expensive); but `std::vector` is freed when it goes out of scope (that's C++). I ran the test code with the fix, the memory is kept around 40M~50M on OS X after running for at least 1 minutes.

Pzixel: @nebgnahz done

Pzixel:
@nebgnahz ping? 🛩

nebgnahz:
Merged, thanks!

PR #32, opened 2018-01-07
Instruction fix

Pzixel:
I tried to follow my own instruction and have found that it's a bit incomplete. So this guy fixes it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ In this case, configuration process is a bit more complicated, because you have 
 - Click ‘Generate’. Select 'MinGW makefiles' from the dropdown and click ok. Now you have properly configured OpenCV.
 - Open bash shell in `OPENCV_DIR%\x64\mingw64` directory and run `mingw32-make`. You should get following result:
 ![MinGW make](https://preview.ibb.co/fuTz1b/2018_01_07.png)
-- Set environment variable `%OPENCV_LIB%` to `%OPENCV_DIR%\x64\mingw64\lib`, which has `opencv_world<version_number>.dll.a`.
+- Set environment variable `%OPENCV_LIB%` to `%OPENCV_DIR%\x64\mingw64\lib`, which has `libopencv_world<version_number>.dll.a`.
 
 
 ## Usage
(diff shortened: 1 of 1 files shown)


Pzixel:
I also propose squash these commits. And I have another problem: for some reason it doesn't work on my another PC :) I checked my configuration twice but I see no difference. I have set both environment variables, I added `%OPENCV_DIR%\bin` to the PATH but when I run `cargo test` I just get following error:
```
   Compiling cv v0.1.0 (file:///C:/Users/Alex/Documents/Rust/cv-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 7.21 secs
     Running target\debug\deps\test_features2d-e23864be15c5cd5e.exe
error: test failed, to rerun pass '--test test_features2d'

Process finished with exit code -1073741515 (0xC0000135)
```

without any details.

Any ideas? I also propose to merge this guy when we solve this problem. We probably need add something in this instruction.

nebgnahz:
Squashing these commits is a great idea. Trick that I know is either `rebase` or reset softly, i.e., `git reset --soft "HEAD~n"` and then `git commit --amend`.

I don't know exactly what's going on with the test case. `0xC0000135` on Windows seems to be something related with `.dll`, so the library is not properly configured? Do other tests pass? Try `cargo test -v` and maybe debug with `cargo test -- --nocapture` to see what's going on.

Pzixel:
@nebgnahz yeah, my instruction is actually wrong. Going to fix it now.

Pzixel:
@nebgnahz fixed

nebgnahz:
Ok, I've merged. Thanks.

P.S. I don't think Github does squashing.

Two changes that I am thinking:
1. I would prefer to host the screenshot on Github.
2. I may move the installation to a separate `Installation.md` file to clean the clutter in README.

nebgnahz:
Screenshot host link: [mingw32-make](https://user-images.githubusercontent.com/941519/34654809-836fe56c-f3b5-11e7-87b3-46bb7667bf34.png)

Pzixel:
@nebgnahz 
image

nebgnahz:
Ah, I missed the drop down icon when clicking the `merge` button... Lesson learned today!

And I agree with your argument that a separate `Installation.md` will make it looks complicated (which I think it is, on Windows, sort of).

Pzixel:
@nebgnahz no, it's just because opencv is precompiled for MSVC and you have to compile GNU version yourself. It's not really complicated, it's just detailed. I could just write `cmake all the stuff`, but it's may be not familiar for windows users, most of them are familiar with C# which has as high level depenency model as rust has: package manager and so on. CMAKE is too low level, so I just explained all step in details to make it easy to perform for newcomers too.

PR #34, opened 2018-01-09
Add GPU features in test suite

Pzixel:
-  Travis:  `cargo:warning=native/opencv-gpu.cc:4:37: fatal error: opencv2/cudaobjdetect.hpp: No such file or directory` 
- Appveyor: `native/opencv-gpu.cc(4): fatal error C1083: Cannot open include file: 'opencv2/cudaobjdetect.hpp': No such file or directory`

it seems that GPU feature is completely broken on both Linux and Windows :)


Pzixel:
cargo:warning=native/opencv-gpu.cc:4:37: fatal error: opencv2/cudaobjdetect.hpp: No such file or directory - it seems that GPU feature is totally broken

Pzixel:
`cudaobject.hpp` is located in `opencv\sources\modules\cudaobjdetect\include\opencv2`, instead of `opencv\build\include`. We don't pass this path to the compiler so it can't find it:
```bash
$ g++.exe -O0 -ffunction-sections -fdata-sections -g -m64 -I native -I C:\\opencv\\build\\include --std=c++11 -Wall -Wextra -o C:\\Users\\SOLO\\Documents\\Rust\\cv-rs\\target\\debug\\build\\cv-a0389601abcc4f40\\out\\native\\opencv-gpu.o -c native/opencv-gpu.cc
native/opencv-gpu.cc:4:10: fatal error: opencv2/cudaobjdetect.hpp: No such file or directory
 #include <opencv2/cudaobjdetect.hpp>
```

nebgnahz:
GPU features are not broken; they are just not properly set up on CI platforms. I don't think Travis free tier can do it.

I am not sure directly referencing `source` is a good idea. Even if you can compile by finding the `.h` files (after you downloaded the source), you still need to link to make the library work. An installation is necessary (and after you install, you should have `include` directory).

The reason that `opencv2/cudaobjdetect.hpp` is not available might be due to a CMake flag `WITH_CUDA=OFF` (only guessing here).

Pzixel:
Anyway, we need all tests being passing. I can try to tweak windows version, but I haven't linux environment right now to fix it too.

Pzixel:
Okay, maybe I was too rude about said `completely broken`, but it's unusable. I'm playing with CMake options. One of them just doesn't work (`BUILD_CUDA_STUBS`). I think that we have to fix this problem anyway, because otherwise it's just like "configure your system somehow. It it works, let us know how you did it" :)

Pzixel:
Okay, it's not possible to use MinGW for CUDA. Very nice..
```c
if(WIN32 AND NOT MSVC)
  message(STATUS "CUDA compilation is disabled (due to only Visual Studio compiler supported on your platform).")
  return()
endif()
```
Going to try with MSVC...

nebgnahz:
Travis can build cuda/GPU code but I don't think you can run the test (not in N2D2) and [this issue from Travis](https://github.com/travis-ci/travis-ci/issues/5911) is still open.

Thanks for the pointer to N2D2 scripts. I will look into adding Linux build support on Travis tomorrow.

Pzixel:
@nebgnahz ok, I see. However, AppVeyor seems to be supporting this. I believe script for N2D2 is taken from here: https://github.com/willyd/appveyor-cuda-test

Pzixel:
Yet another tip: we may use `#[cfg(windows, target_env = "msvc")]` for GPU code to make sure that people will get a compile time error for incompatible toolchain. https://doc.rust-lang.org/beta/reference/attributes.html#conditional-compilation

nebgnahz:
I tried to build OpenCV with CUDA support from source on Travis, it took so long that Travis killed the job.

On travis-ci.org, it stopped after compiling for 49 min 18 sec.
On travis-ci.com, it stopped after compiling for 1 hr 59 min 26 sec.

Pzixel:
@nebgnahz yep, I have similar results on my local machine: I started build 1 hour ago, I guess, and it still compiling... But we need test GPU features otherwise it's just untested. "It works locally" never really works in real life :)

nebgnahz:
One way is to use Docker to package all dependencies and test the code.

Pzixel:
@nebgnahz you won't be able to use docker for windows because windows docker images has size about 4gb... Sad but true.

nebgnahz:
Alright, thanks for the link for 6x speed up. I do see substantial improvements and now the build is passing (see MR #36).

A couple of additional changes:

1. create `.ci` folder so that any build script for AppVeyor can go there.
2. fix `CVec` bug in `cuda.rs`

Pzixel:
@nebgnahz wonderful job. One more step in the right direction :)

nebgnahz:
Oh, I didn't mean to move `AppVeyor.yml` or `.travis.yml` to `.ci` folder. I thought the folder is a good place to put scripts like shell scripts (for *nix), or cmd scripts (for Windows). Alternatively, we can create a `scripts` folder.

Pzixel:
@nebgnahz oh, I misread this one. No, this is fine as it is.

Pzixel:
If you wonder, I'm not dissapeared 😄  I'm stuck with creating `include` folder on windows. It exists in prebuild `vc14_cv15` archive but I have no idea how populate it.

Pzixel:
I'd like to report some progress. I was able to reproduce an official build (with `include` folder and so on) by downloading repo and compiling it locally: https://docs.opencv.org/master/d3/d52/tutorial_windows_install.html#tutorial_windows_gitbash_build .

Now we just have to adjust some switches and we're golden.

nebgnahz:
Alright, as someone who has struggled with Linux/OS X cuda installation and someone who stays away from Windows because it's not developer friendly, I really appreciate your effort here! Thanks!

Pzixel:
@nebgnahz 

>  because it's not developer friendly

That's funny, but it seems that build is frozen because it's showing UI popup message :) And I have permissions to cancel it. So cancel it, and probably give me `Run project builds` role, because I have to play a lot with CI.

Commit 2be7b17 is ok, it can build OCV from scratch, now I'm trying to enable CUDA.

Pzixel:
Got tired a bit. Going to continue tomorrow. For unknown reason, after successful installation of CUDA there is no `nvcc` cuda compiler available. It probably fails for some reason, but do it silently without any evidences...

Pzixel:
It seems that I found the correct combination of flags etc, but compilation exeeds 1h and terminates by timeout. I already used "specify architecture explicitely" workaround, but it's not enough.

nebgnahz:
Hey, @Pzixel, I've configured AppVeyor Permission, now users are allowed to `run build` and `update settings`. Is that what you mean?

Also I configured the remote desktop password on AppVeyor and sent the email containing password to you (gmail) privately. It should allow you to have remote desktop access (should have done this earlier to save the debug time).

Do you need to include `opencv_contrib`? I know it's something good to have, but that will increase the build time for sure. Maybe skip it first to reduce the build time?

Pzixel:
@nebgnahz hi again! Thank you for permissions, I'l use it if it's necessary. 

> Do you need to include opencv_contrib? I know it's something good to have, but that will increase the build time for sure. Maybe skip it first to reduce the build time?

I'm not sure about it, it's how default script have been written. I considered that it's probably won't increase build time significally, because it's downloaded and built pretty quick. You can see that we had a build with installed CUDA and manually compiled OpenCV ([link](https://ci.appveyor.com/project/nebgnahz/cv-rs/build/1.0.154)). It took 40 minutes to complete. So CUDA adds more than 20 minutes here.

And don't worry, we can cache all this stuff, so we won't spend an hour every time we compile the project.

Tomorrow I'l try to rewrite build script to speedup repository download and remove `opencv-contrib` from switch. It would probably allow us to be in time.

Pzixel:
Another question: do we need `opencv_world340d` file? Because AppVeyor cache is limited to 1GB, while these two files are 1.7GB on my machine...

image

nebgnahz:
I think it's safe to remove it. My understanding is that `opencv_world<version>.dll` is the release mode build, while `opencv_world<version>d.dll` is the debug build (i.e. with debug symbols).

Pzixel:
Hey, it's working! :) Tests are failing, but it's building at least. https://ci.appveyor.com/project/nebgnahz/cv-rs/build/1.0.175

Pzixel:
Unfortunately, I don't see any button for strating or cancelling builds :)

nebgnahz:
Great progress. I guess I might need to add you as Github repo collaborator (just sent the invitation).

I don't have much experience with the collaborator functionality in AppVeyor.

Pzixel:
Glad to hear it :)

Well, windows configuration is almost done. I will squash these 40-or-how-many-i-already-forgot commits when it is completely done.

For now I have again this issue: https://stackoverflow.com/questions/45918446/compiling-opencv-3-3-with-cuda-9-0rc.

Have no clue why it is happening becasue it was building 5 hours ago and I didn't change configuration... Maybe it's `previus VS` issue, so we probably have to switch to 2015 image...

Pzixel:
Wow, it's alive: https://ci.appveyor.com/project/nebgnahz/cv-rs/build/1.0.192

Tests are not passing, not even with `--no-default-features`, but it builds, at least :)

I hope tomorrow I'l fix an issue with cache, and therefore we will enhance our build time

Pzixel:
I added some changes to `yml` files. Now, AppVeyor won't start build if you change `sh` files or `travis.yml`. It doesn't work vice verca, so I just exluced this branch and every branch starting with name `win` from travis builds. So it shouldn't be triggered anymore and just waste CI server time.

Pzixel:
@nebgnahz I need your help :)

It builds, but for `cargo test` it fails with multiple errors such as:

> C:\Users\Alex\Documents\Rust\cv-rs\target\debug\deps\libcv-915e0e48e55af360.rlib(opencv-wrapper.o): In function `cv::String::String(char const*)':
>           C:/Users/Alex/Documents/foo/install/opencv/include/opencv2/core/cvstd.hpp:602: undefined reference to `cv::String::allocate(unsigned long long)'

I have no idea how to fix them. Probably some modules didn't install? However, `String::allocate` seems to be very basic, I don't know why it says it cannot link it... Any ideas?

nebgnahz:
`cv::String::allocate` seems to be the first thing that OpenCV looks for when it runs; so this is just a sign of "failed to link against the right library" (see the similar issue here when we were fixing OpenCV windows build script: https://github.com/nebgnahz/cv-rs/issues/26#issuecomment-355084502).

Does this happen on your machine (looks like so by "C:\Users\Alex")? Maybe check `opencv_world<version>.dll`? Does this also happen on AppVeyor?

Pzixel:
@nebgnahz well, dll defenitly exists, as well as `include` folder.

You can create a VM with Windows yourself if you want, then just use trial, it's doesn't limited in any functionallity. AppVeyor script stepsa allow you to reproduce very this situation on any machine because it's supposed to configure clean machine. I'l try to do it on my laptop instead of desktop PC, maybe i'l learn something.

Pzixel:
I also propose to enable rolling builds to cancel build if there is new commit in the same branch:

image001

nebgnahz:
Just configured with rolling builds and saving caches for PR.

screenshot 2018-01-18 10 52 44

I will try to find a time and test out this on Windows...

Pzixel:
I also reproduced weird test running locally
image
Going to investigate...

Pzixel:
Current error is `nvcuda.dll is missing` on CI server. Continuing investigation..

Pzixel:
Here here, my final results:

AppVeyor (and Travis too, I guess) doesn't provide CUDA-compatible VMs so it's not possible to run GPU feature tests there. 

Thus The best we can do is provide script for a users to setup CUDA from scratch: current `appveyor_build_opencv` and `appveyor_install_cuda` are well written and tested.

Unfortunately, this PR should be closed (and we will remember these 57 commits full of pain) and I'l open a new one with enhanced `appveyor.yml` and user scripts, based on achieved experience.

nebgnahz:
Alright, thanks for the perseverance in enabling GPU build on Windows/AppVeyor.

Does the current build pass on AppVeyor? I see you mentioned `nvcuda.dll` is missing, has that been fixed?

Apologies that my schedule is too packed recently...

Pzixel:
@nebgnahz `nvcuda.dll is missing` means that installation didn't find nVidia GPU and therefore didn't complete installation. And it's clear why it's happening: VMs has no physical GPU, they have `HyperV Standard GPU` device or something similar in Travis case.

Generally speaking, there is no way to run CUDA on Travis or AppVeyor. The best we can do is build and check that wrappers are written correctly. 

Because we can check it on Travis **or** AppVeyor, I suppose we won't add these changes to the upstream. I'l create a new PR with script `click_here_and_I_will_configure_opencv_with_cuda_for_you.ps1` to make it easier for newcomers to install it if they have compatible hardware. But we have no chance to run in on CIs 😄

nebgnahz:
Sounds great to me.

Why don't we open issues for these roadmap items you have so that we can track each one of them?

I am closing this PR as you have #38 prepared.

PR #41, opened 2018-01-21
Get rid of libc

Pzixel:
We have no need in `libc` because we must use `std::os::raw` and `std::ffi` instead.


Pzixel:
@nebgnahz I actually think something like 👍 would be better 😄

PR #45, opened 2018-01-22
Documentation fix

Pzixel:
Here here docs fix (I'm done with all warnings) + I have found a solution for our examples that doesn't require any changes: just move all related dependencies in `dev-dependencies` section:

> Dev-dependencies are not used when compiling a package for building, but are used for compiling tests, **_examples_**, and benchmarks.

I have also played with appveyor cache, it gives us 20 seconds win. Well, why not?:)

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -2,18 +2,18 @@
 use std::path::PathBuf;
 
 #[derive(Debug, Fail)]
-///Custom errors that may happen during calls
+/// Custom errors that may happen during calls
 pub enum CvError {
     #[fail(display = "invalid path: {:?}", path)]
-    ///Indicates that path was invalid
+    /// Indicates that path was invalid
     InvalidPath {
-        ///Path that caused an error
+        /// Path that caused an error
         path: PathBuf,
     },
     #[fail(display = "failed to convert from primitive: {}", value)]
-    ///Indicates that conversion from primitive to enum type is failed
+    /// Indicates that conversion from primitive to enum type is failed
     EnumFromPrimitiveConversionError {
-        ///Value that caused an error
+        /// Value that caused an error
         value: i32,
     },
 }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -8,12 +8,12 @@ pub enum CvError {
     ///Indicates that path was invalid
     InvalidPath {
         ///Path that caused an error
-        path: PathBuf
+        path: PathBuf,
     },
     #[fail(display = "failed to convert from primitive: {}", value)]
     ///Indicates that conversion from primitive to enum type is failed
     EnumFromPrimitiveConversionError {
         ///Value that caused an error
-        value: i32
+        value: i32,
     },
 }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,10 +1,19 @@
 //! Errors for OpenCV bindings
-#![allow(missing_docs)]
-
 use std::path::PathBuf;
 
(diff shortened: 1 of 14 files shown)


src/errors.rs
```diff
@@ -1,10 +1,19 @@
 //! Errors for OpenCV bindings
-#![allow(missing_docs)]
-
 use std::path::PathBuf;
 
 #[derive(Debug, Fail)]
+///Custom errors that may happen during calls
```
nebgnahz: I think this should trigger `cargo fmt`.

src/videoio.rs
```diff
@@ -220,33 +220,33 @@ extern "C" {
...
-    /// * fourcc – 4-character code of codec used to compress the frames. For
-    ///   example, VideoWriter::fourcc('P','I','M','1') is a MPEG-1 codec,
-    ///   VideoWriter::fourcc('M','J','P','G') is a motion-jpeg codec etc. List
-    ///   of codes can be obtained at Video Codecs by FOURCC page.
-    /// * fps – Framerate of the created video stream.
-    /// * frame_size – Size of the video frames.
-    /// * is_color – If it is not zero, the encoder will expect and encode color
-    ///   frames, otherwise it will work with grayscale frames (the flag is
-    ///   currently supported on Windows only).
+    /// -path – Name of the output video file.
```
nebgnahz: Is this fix suggested from the new cargo doc pulldown-cmark? I am surprised because removing the space reduces readability, IMHO.

Pzixel: Yep, see [this link](https://blog.guillaume-gomez.fr/articles/2017-09-18+New+rustdoc+rendering+common+errors), about lists

src/errors.rs
Pzixel: It doesn't. However, I fixed it. I did it because of lists issue, I expected that all comments should not have an empty line before text. It's not an issue so I turn it back.

nebgnahz:
The test failing cases are not relevant. I am ok with merging this and address the `FromPrimitive` cast later.

Pzixel:
It seems that it doesn't build, and it's funny that I don't understand how's this possible.

Consider this screenshot:
image
We have a successful build in changset *1*, then we have some changeset *2* which breaks the build. Ok, we revert it and get the changeset *3* which should be identical to *1*, but for some reason it fails...

I have no idea what's wrong here, except that environments changed somehow.

Pzixel:
I think I found what's causing this problem 😄 
image

nebgnahz:
I think trailing comma is actually fine (no strong opinion).

We can check code format on Travis (optionally AppVeyor as well).
```cargo fmt -- --write-mode=diff``` and it will return `4` (non-zero) if the code is not formatted.

**Comment by Pzixel** [comment]
I mean I don't really like `trailing comma` rule, however, it's default, [see documentation](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#trailing_comma).
If you like it and it's default, then 2 points for it.
If I dislike it then 1 point against.

Result - 2:1, so we stay with this default.


P.S. FYI, `cargo fmt` is not currently working on stable toolchain, because for some reason it isn't added to `cargo`. However, `rustfmt` itself exist. So my current workaround on windows is:
```powershell
Get-ChildItem -File -Filter *.rs -Recurse | % { &"C:\Users\Alex\.rustup\toolchains\stable-x86_64-pc-windows-gnu\bin\rustfmt.exe" $_.FullName }
```
You may use similar workaround on linux.

PR #46, opened 2018-01-24
check rustfmt on travis

nebgnahz:
For now, we use `rustfmt-preview` and find/xargs to run rustfmt.
Once `cargo fmt` is available in stable, we will switch to use `cargo`.

References: #45 

@Pzixel ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: trusty
 language: rust
 rust:
   - stable
+  - nightly
 
 addons:
   apt:
@@ -38,6 +39,9 @@ script:
   - cargo build --features gpu
   - cargo doc --features gpu --no-deps
   - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo bench ; fi
+  - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
+      cargo fmt -- --write-mode=diff;
+    fi
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,17 +27,21 @@ addons:
...
 
 script:
   - cargo build
   - cargo build --no-default-features
   - cargo test --no-default-features
   - cargo build --features gpu
   - cargo doc --features gpu --no-deps
-  - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo bench ; fi
+  - find src/*.rs -exec rustfmt {} \;
+  # Available until `cargo fmt` landed in stable rust 1.24
```
Pzixel: As I said, you don't need `cargo fmt` here, just run `rustfmt` manually from `.cargo` directory.

.travis.yml
```diff
@@ -27,17 +27,21 @@ addons:
...
+  # - cargo fmt -- --write-mode=diff;
```
Pzixel: > error: toolchain 'stable-x86_64-unknown-linux-gnu' does not have the binary `rustfmt`

You have to install it before

nebgnahz: You are right. I mis-understood the return code of `which rustfmt`. `which` succeeds but it's not usable.

Pzixel:
👍

nebgnahz:
Yes, [it catches](https://travis-ci.org/nebgnahz/cv-rs/builds/333137080).

PR #47, opened 2018-01-25
compareHist

Pzixel:
I have found that templates are OK in FFI interop.

I think we should convert `VecType` macro to the template, becase it's shows better its intent and is easier to support.

---

Hmm, interesting. It does work with gcc, but it doesn't with msvc. Very bad :(


Pzixel:
We have inconsistency with OCV versions. We had 3.3.0 on linux and 3.4.0 on windows. It leads to different results in several methods, so tests are passing on 3.4.0, but fails for 3.3.0. I updated `travis.yml` to compile against 3.4.0, but it fails because of timeout. I think you have to ask support to elevate timeout for the project like I did for `appveyor`.

src/lib.rs
```diff
@@ -8,8 +8,6 @@
 //! [opencv-rust](https://github.com/kali/opencv-rust/) which generates OpenCV
 //! bindings using a Python script.
 #![deny(missing_docs)]
-#![deny(missing_debug_implementations)]
-#![deny(missing_copy_implementations)]
```
Pzixel: I'm not sure about these denies. I removed them when I added global `Cv` struct without any fields, just to create bunch of static methods. Now they can be restored, but i'm not sure if they should.

nebgnahz:
Looks like [build #261](https://travis-ci.org/nebgnahz/cv-rs/builds/333739835) uses 3.4.0 failed but not timed out. Am I reading this  correctly?

Pzixel:
@nebgnahz yep, previous build failed due to timeout, but this one actually failed because of tests... That's weird, because they are passing on AppVeyor. Can you run them yourself on your machine? Because I don't understand why we get different results on different environments.

nebgnahz:
I was able to reproduce the assertion failure on my machine (Opencv 3.4.0, Linux Ubuntu 17.10):

```
cargo test --test test_imgproc -- --nocapture
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running target/debug/deps/test_imgproc-24b3d58bcb1e766c

running 7 tests
OpenCV Error: Assertion failed (A.size == arrays[i0]->size) in init, file /home/benzh/Downloads/temp/opencv-3.4.0/modules/core/src/matrix.cpp, line 4845
thread 'compare_hist_different_dimensions_panic' panicked at 'called `Result::unwrap()` on an `Err` value: "/home/benzh/Downloads/temp/opencv-3.4.0/modules/core/src/matrix.cpp:4845: error: (-215) A.size == arrays[i0]->size in function init\n"', /checkout/src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
test compare_hist_different_dimensions_panic ... ok
thread 'thread 'compare_hist_bhattacharyya' panicked at '0.6798259690477988 == 0.6679', tests/test_imgproc.rs:81:4
compare_hist_intersection' panicked at 'test compare_hist_bhattacharyya ... FAILED
5.440850785933435 == 5.682', tests/test_imgproc.rs:81:4
test compare_hist_intersection ... FAILED
thread 'compare_hist_chi_square_alternative' panicked at '41.65613074156445 == 41.027', tests/test_imgproc.rs:81:4
test compare_hist_chi_square_alternative ... FAILED
thread 'compare_hist_kullback_leibler_divergence' panicked at '55.71912075710992 == 54.06287', tests/test_imgproc.rs:81:4
test compare_hist_kullback_leibler_divergence ... FAILED
thread 'compare_hist_correlation' panicked at '0.20456957916644988 == 0.211', tests/test_imgproc.rs:81:4
test compare_hist_correlation ... FAILED
thread 'compare_hist_chi_square' panicked at '2697.981816285168 == 1360.7', tests/test_imgproc.rs:81:4
test compare_hist_chi_square ... FAILED

failures:

failures:
    compare_hist_bhattacharyya
    compare_hist_chi_square
    compare_hist_chi_square_alternative
    compare_hist_correlation
    compare_hist_intersection
    compare_hist_kullback_leibler_divergence

test result: FAILED. 1 passed; 6 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test test_imgproc'
```

nebgnahz:
5 tests are due to float number comparisons; one is OpenCV assertion failure (the `should_panic` one).

Pzixel:
@nebgnahz but they are passing on AppVeyor. It means values returned by OpenCV on windows differ from linux ones.

nebgnahz:
My machine has the same results with Travis'. So yes, the results seem to be OS dependent. My Mac is currently using OpenCV 3.3 so I cannot test it on Mac at this point (I suspect the results will also be different).

Pzixel:
[This question](https://stackoverflow.com/questions/40065011/opencv-behavior-differences-between-linux-and-windows-built-application) shows that it's probably because of `jpg` image. We may try `bmp` to see if it's image codec who is wrong.

Pzixel:
Anyway, what do you think about changes?

src/imgproc.rs
```diff
@@ -54,6 +56,25 @@ extern "C" {
...
     );
+
+    fn cv_compare_hist(first_image: *const CMat, second_image: *const CMat, method: c_int) -> CResult<c_double>;
+}
+
+/// Possible methods for histogram comparision method
+#[derive(Debug, PartialEq, Clone, Copy, ToPrimitive)]
+pub enum HistogramComparisionMethod {
+    /// HISTCMP_CORREL
+    Corellation = 0,
```
nebgnahz: s/Corellation/Correlation

src/lib.rs
nebgnahz: I added these lints after reading [Good Practices for Writing Rust Libraries](https://pascalhertleif.de/artikel/good-practices-for-writing-rust-libraries/).

Users of this library can compose new types and expect debug print works whenever they should.

src/lib.rs
```diff
@@ -46,3 +46,38 @@ pub mod features2d;
 
 #[cfg(feature = "gpu")]
 pub mod cuda;
+
+extern "C" {
+    fn c_drop(value: *mut c_void);
+}
+
+/// Structure that contains global cv functions
+pub struct Cv {}
```
nebgnahz: Can you clarify what `Cv` here is useful for?

Pzixel: It must be removed

src/imgproc.rs
Pzixel: Let's do it this way: say when you're done with review and I fix it in one commit :)

Pzixel:
Beside the point of this PR: I suggest to rename enums to be more readable. For example, `NormTypes::NormMinMax` could be `Norm::MinMax`, `ImreadModes::ImreadColor` -> `ImageReadMode::Color` and so on. Note the singular form of enum value because it's considered to be better practice, e.g. see [MS docs](https://msdn.microsoft.com/en-us/library/4x252001(v=vs.71).aspx), [Oracle docs](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) or [rust docs themselfs](https://rustbyexample.com/custom_types/enum.html)

src/imgproc.rs
nebgnahz: This is it. I thought when I am done with a review, Github will notify you (I need to click a button says something like "you have to address these comments").

Pzixel: @nebgnahz I imagine there should be a button `I'm done with review` which sends all comment at once. I think I saw it when I was on Github enterprise. I didn't review anybody on github myself so I just don't know :)

Pzixel:
Are you still not agreed with changes?:)
image

nebgnahz:
Oops, forget to change that while I've given "thumb up".

Pzixel:
You did it before your last edit and I assumed that it was related to renaming proposal.

PR #48, opened 2018-01-26
Try to get rid of CVec and CResult macros

Pzixel:
They are really annoying to write. Trying to replace them with templates that can cross FFI boundary.


Pzixel:
Haha, look, it's working! We actually are using templates on C++ side and generics on Rust one. No more Macro `Vec(double, VecDouble)`! 😄

Pzixel:
I removed CVec and CDouble, and I've got hit by idea: why stop here? We could get rid of all these `reinterpret_cast`!

nebgnahz:
Wow, getting rid of `reinterpret_cast` sound great!

Personally I think C++ compiler is a safe assumption; the only reason we are dealing with manual C headers is due to the poor Rust/C++ interop.

Pzixel:
@nebgnahz ok, because I paused development because I wasn't sure if we are one the same wave. Then, let's see what can I do. You can look on changeset at its current state, it's already promising :)

Pzixel:
I think i'm done. As last effort I removed bunch of `vec_xxx_cxx_to_c` methods and replaced it by single `cv_to_ffi` with different overloads. Want to pass something to Rust? Just call this method. And yes, no `reinterpret_calls` left in repository, except for `cv::Mat`' constructor, but it's unavoidable there.

Pzixel:
@nebgnahz  Ping?

native/utils.h
```diff
@@ -9,8 +9,33 @@
 //   Utils
 // =============================================================================
 
-void vec_rect_cxx_to_c(const std::vector<cv::Rect>& cxx_vec_rect, VecRect* vr);
-void vec_double_cxx_to_c(const std::vector<double>& cxx_vec, VecDouble* v);
-void vec_point_cxx_to_c(const std::vector<cv::Point>& cxx_vec_point, VecPoint* vp);
-void vec_points_cxx_to_c(const std::vector<std::vector<cv::Point>> &cxx_vec_points, VecPoints* vps);
+void cv_to_ffi(const cv::Rect& source, Rect* dest){
```
nebgnahz: The implementation should be moved to `.cc` file. Otherwise, when `opencv-wrapper.c` includes `utils.h` and `opencv-gpu.cc` includes `utils.h`, we have a multiple definition issue.
That is, both `opencv-wrapper.o` and `opencv-gpu.o` have definitions of `cv_to_ffi` for the same argument.

Pzixel: Yep, I tried it. However, templates cannot be moved out of `.h` file, see explanation: https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file.
You can add a commit yourself if you see how it can be done.

nebgnahz: Leave the template function in `.h` and move the implementation to `.cc`.

```
// utils.h
void cv_to_ffi(const cv::Rect& source, Rect* dest);
void cv_to_ffi(const cv::Point& source, Point2i* dest);
void cv_to_ffi(const std::vector<double>& source, CVec<double>* dest);

template <typename T, typename U>
void cv_to_ffi(const std::vector<T>& source, CVec<U>* dest)
{
    size_t num = source.size();
    dest->size = num;
    dest->array = (U*) malloc(num * sizeof(U));
    for (size_t i = 0; i < num; i++) {
        cv_to_ffi(source[i], &dest->array[i]);
    }
}

#endif  // UTILS_H_
```

nebgnahz: Mmm, hitting an error:

```
     Running target/debug/deps/test_objdetect-af8e9348272df2ef

running 2 tests
error: An unknown error occurred
```

Let me take a look.

Pzixel: I did it, however I have a question: why double definition is not a problem for templates?

Pzixel: Interesting, it's passing on Travis.

native/opencv-gpu.cc
```diff
@@ -177,102 +151,83 @@ Size2i cv_gpu_hog_get_win_stride(GpuHog* hog) {
 // =============================================================================
 //   CascadeClassifier
 // =============================================================================
-using GpuCascadePtr = cv::Ptr<cv::cuda::CascadeClassifier>;
-
-GpuCascade* cv_gpu_cascade_new(const char* const filename) {
-    auto cascade = cv::cuda::CascadeClassifier::create(filename);
-    return reinterpret_cast<GpuCascade*>(new GpuCascadePtr(cascade));
+void* cv_gpu_cascade_new(const char* const filename) {
+    return cv::cuda::CascadeClassifier::create(filename);
```
nebgnahz: Missing a new here, i.e. `new cv::Ptr<cv::cuda::CascadeClassifier>`

native/utils.h
nebgnahz: I traced down the bug. See my updated comment (request for change).

native/opencv-gpu.cc
Pzixel: Yes, I see.

Why travis didn't handle it?

nebgnahz:
The bug triggers my thought: because the current code does implicit conversion, the compiler won't warn us for type miss match because we are casting everything to void*. Maybe leave some `reinterpret_cast` and those custom C types?

native/utils.h
nebgnahz: templates are not compiled to object files, see https://stackoverflow.com/questions/7584781/inclusion-of-header-files-in-case-of-templates

Pzixel:
I was thinking about it for a while, and I decided that's it's ok because 
1. We only use `void*` as return type, in several places, typically constructors only
2. `reinterpret_cast` doesn't guarantee anything, so it's just a false promise that types matches. It's just like `service locator`, if you know this pattern. It seems that it's statically typed and it's fine but it isn't.

That's what I think about it.

nebgnahz:
I see your point and agree that the `reinterpret_cast` has no guarantee. These constructor functions are short, and hopefully we can easily trace down bugs in the future.

P.S. I should have used `static_cast` from early on instead of `reinterpret_cast`.

native/common.h
```diff
@@ -0,0 +1,81 @@
...
+            auto len = std::strlen(err_msg);
+            error = new char[len + 1];
+            std::strcpy(error, err_msg);
+        }
+        return Result<T>{value, error};
+    }
+};
+
+template<typename T>
+struct CVec
```
nebgnahz: `CVec {`?

I am mostly following the default LLVM C/C++ styles, see my `.clang-format` as below:

```
BasedOnStyle: llvm
IndentWidth: 4
AllowShortFunctionsOnASingleLine: None
DerivePointerAlignment: false
PointerAlignment: Left
KeepEmptyLinesAtTheStartOfBlocks: false
SpaceAfterCStyleCast: true
SpacesBeforeTrailingComments: 2
AlignAfterOpenBracket: Align
IndentCaseLabels: false
```

nebgnahz: OK!

Issue #40, opened 2018-01-21
Roadmap 21.01.2018


Pzixel:
- [x] Create user script to enable easy OCV+CUDA configuration (#38)
- [x] Add `.gitattributes` (#38)
- [x] Enable rustfmt to allow better review and single source of truth in questions of how code should be written. (#39)
- [x] Get rid of libc, because of multiple compatibility issues. (#41)
- [x] Create user script to enable easy OCV+MinGW configuration (#42)
- [x] Get rid of `errors_chain` crate and use `failure` one ([see explanation](https://www.reddit.com/r/rust/comments/7h8v1z/errorchain_and_failure/)) (#44)
- [x] Fix cargo doc issues (#45)
- [x] Add more features such as cv::compareHist. (#47)
- [x] Get rid of macros such as `CVec`, `CResult` and so on and try to use templates. I have some ideas here. (#48)
- [x] Format C++ code. (#53)

Pzixel:
Another idea is move examples to a separated project because otherwise you have dependencies (such as `getopts`) that library actually doesn't use.

nebgnahz:
We can factor examples out and have each example become a standalone Cargo project.

Pzixel:
Yep, that's what I mean

Pzixel:
Wow, it's been 3 days and we already did almost all from the list.

Pzixel:
@nebgnahz any progress in finding `cppfmt` for C code (last bullet of the list)? I recall that you was expecting code that is formatted in single manner.

PR #53, opened 2018-02-04
clang-format and check format on Travis

nebgnahz:
- Check in .clang-format spec (based on LLVM)
- The source code in `native` has been formatted with `clang-format -i native/*`.
- Travis checks for clang-format diffs.
- Update `CONTRIBUTE.md` to include clang format check.

References: #40 

@Pzixel I am done with the clang format for now. So this is ready to review. Leave any style suggestions, or other implementation comments.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,94 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: false
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     80
+CommentPragmas:  '^ IWYU pragma:'
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeCategories:
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
+    Priority:        2
+  - Regex:           '^(<|"(gtest|isl|json)/)'
+    Priority:        3
+  - Regex:           '.*'
+    Priority:        1
+IncludeIsMainRegex: '$'
+IndentCaseLabels: false
+IndentWidth:     4
+IndentWrappedFunctionNames: false
+JavaScriptQuotes: Leave
+JavaScriptWrapImports: true
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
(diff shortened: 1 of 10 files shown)


native/opencv-gpu.cc
```diff
@@ -171,63 +178,76 @@ void cv_gpu_cascade_detect(cv::Ptr<cv::cuda::CascadeClassifier>* cascade, cv::cu
...
+    cv::Ptr<cv::cuda::CascadeClassifier>* cascade, double factor) {
     (*cascade)->setScaleFactor(factor);
 }
 
-Size2i cv_gpu_cascade_get_classifier_size(cv::Ptr<cv::cuda::CascadeClassifier>* cascade) {
+Size2i cv_gpu_cascade_get_classifier_size(
+    cv::Ptr<cv::cuda::CascadeClassifier>* cascade) {
     cv::Size2i size = (*cascade)->getClassifierSize();
-    Size2i c_size = {.width = size.width, .height = size.height };
+    Size2i c_size = {.width = size.width, .height = size.height};
```
Pzixel: This syntax is GCC-only, it won't work on MSVC. We don't compile GPU modules on AppVeyor, this is why this thing happen. I think it should be fixed. Correct syntax for both compilers is `Size2i c_size = {size.width, size.height};`. I had the same issue with `CVec`.

nebgnahz: Thanks for catching this. Fixed.

Pzixel: @nebgnahz No problem. I think there are some other places too. Lines 202, 220, 226.

Pzixel:
Consider adding information into `Contribute.md` about formatting. People may be confused why build doesn't pass on CI.

native/opencv-gpu.cc
```diff
@@ -1,16 +1,15 @@
...
 
 EXTERN_C_BEGIN
 
 // =============================================================================
 //   Basic
 // =============================================================================
-void* cv_gpu_mat_default() {
-    return new cv::cuda::GpuMat();
-}
+void* cv_gpu_mat_default() { return new cv::cuda::GpuMat(); }
```
Pzixel: Probably we want always use multiline functions?

native/opencv-gpu.cc
```diff
@@ -26,7 +25,7 @@ void* cv_mat_from_gpu_mat(cv::cuda::GpuMat* gpu_image) {
 }
 
 void* cv_gpu_mat_from_mat(cv::Mat* image) {
-  return new cv::cuda::GpuMat(*image);
+    return new cv::cuda::GpuMat(*image);
```
Pzixel: especially when it actually mix single and multiline styles for unknown reason.

.clang-format
```diff
@@ -0,0 +1,94 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        8
```
Pzixel: 8 is probably too bold. Common value is 4 AFAIK

native/opencv-gpu.cc
nebgnahz: Yes, I am tweaking the formatting rules now... Before committing new changes, I want to see Travis is properly configured.

.clang-format
nebgnahz: Tab is never used (see next line). This is just a dump of `.clang-format`.

native/opencv-gpu.cc
nebgnahz: I think this line would exceed 80 if put on a single line. But you are right, the good style should always have multiple lines for this situation.

.travis.yml
```diff
@@ -44,6 +44,9 @@ script:
   - find src/*.rs | xargs rustfmt --write-mode=diff
   # Available until `cargo fmt` landed in stable rust 1.24
   # - cargo fmt -- --write-mode=diff;
+  - |
+    clang-format native/* -output-replacements-xml |
+    grep -c "<replacement " > /dev/null; test $? -eq 1
```
Pzixel: I think it's better to actually write diff if any, it may help to identify problem for people with different `clang-format` settings.

nebgnahz: What do you mean by `write diff`?

Using `clang-format` locally to identify problems with different settings is easier than on Travis...

nebgnahz: I can understand where you are coming from and see how this is useful. `clang-format` has a reasonable good reputation among C/C++ world so I would believe local and CI results would be the same.

P.S. `clang-format` output xml is hard to parse, say, the following example:

```
<replacements xml:space='preserve' incomplete_format='false'>
<replacement offset='16849' length='1'>&#10;</replacement>
</replacements>
```

If we really wanted to show the difference, we might need `clang-format-diff`. I would prefer to add that if the clang format check on CI becomes a problem and people need to debug it repeatedly.

Pzixel: Yes, I see your point.

> If we really wanted to show the difference, we might need clang-format-diff. I would prefer to add that if the clang format check on CI becomes a problem and people need to debug it repeatedly.

Ok, but it seems that adding it is as simple as `diff -u <(cat src/*) <(clang-format src/*)`

Pzixel:
Hmm, what if we just add script that adds git hook on pre-commit, that formats the code via rustfmt and this tool? In this case people won't get tired of `oh, build failed again because I forgot format this single quote. Oh, what the hell". I wrote the following script on my current job, it's really useful:

```
rm .git/hooks/commit-msg
rm .git/hooks/post-checkout
rm .git/hooks/post-merge

echo '#!/usr/bin/env bash
# set this to your active development branch
#develop_branch="develop"
#current_branch="$(git rev-parse --abbrev-ref HEAD)"
#[ "$current_branch" != "$develop_branch" ] && exit 0
commit_regex="(solo-[0-9]+|merge)"
error_msg="Aborting commit. Your commit message is missing either a JIRA Issue ('SOLO-1111') or 'Merge'"
if ! grep -iqE "$commit_regex" "$1"; then
    echo "$error_msg" >&2
    exit 1
fi'  > .git/hooks/commit-msg

echo '#!/bin/sh
./run_gulp.sh ./impl/solo/SolomotoWebRole & ./run_gulp.sh ./impl/solo/Showcases/Solomoto.Showcases.Renderer
wait' | tee -a .git/hooks/post-checkout .git/hooks/post-merge

chmod +x .git/hooks/commit-msg
chmod +x .git/hooks/post-checkout
chmod +x .git/hooks/post-merge

echo "Hooks updated"
```

.travis.yml
nebgnahz: Nice tip! In this way, I also don't need to do the weird grep/test afterwards!

nebgnahz:
I like the idea of adding git hooks. We should host such scripts within the repository and update the README to instruct users about how to set it up. Let's do it in the next PR.

PR #54, opened 2018-02-05
Hooks

Pzixel:
First attempt to add hooks. Don't know it it correctly determines path to `rustfmt` on Linux. Can you check it @nebgnahz, please? 

I'm also waiting for commiting our C rules to make it happen too.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,94 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: false
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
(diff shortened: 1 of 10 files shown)


setup_hooks.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+rustup component add rustfmt-preview
+
+rustfmt_path=`find $HOME/.rustup/toolchains/ -wholename '*bin/rustfmt*'`
```
nebgnahz: `which rustfmt` should suffice?

setup_hooks.sh
```diff
@@ -0,0 +1,10 @@
...
+echo "#!/bin/sh
+find . -name *.rs -type f -print | xargs $rustfmt_path"  > .git/hooks/pre-commit
```
nebgnahz: This only finds *.rs in the top leve.

Also `xargs` will just write all the diffs, and the return code is 0, i.e. the commit will pass.
```
$ git add src/lib.rs
$ git commit -m "small tweaks" // succeed despite format issue
$ git status
Changes not staged for commit:
   modified: src/lib.rs
```

We would want the commit to fail if the format is wrong.

setup_hooks.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
nebgnahz: P.S. for this file as an executable, you should also `chmod +x`.

setup_hooks.sh
Pzixel: On my machine I have following:
```
SOLO@DESKTOP-0LLRSJV MINGW64 ~/Documents/Github/cv-rs (feature/FLANN)
$ find $HOME/.rustup/toolchains/ -wholename '*rustfmt*'
/c/Users/SOLO/.rustup/toolchains/stable-x86_64-pc-windows-gnu/bin/rustfmt.exe
/c/Users/SOLO/.rustup/toolchains/stable-x86_64-pc-windows-gnu/lib/rustlib/manife                         st-rustfmt-preview-x86_64-pc-windows-gnu
/c/Users/SOLO/.rustup/toolchains/stable-x86_64-pc-windows-gnu/share/doc/rustfmt
/c/Users/SOLO/.rustup/toolchains/stable-x86_64-pc-windows-gnu/share/doc/rustfmt/                         LICENSE-APACHE
/c/Users/SOLO/.rustup/toolchains/stable-x86_64-pc-windows-gnu/share/doc/rustfmt/                         LICENSE-MIT
/c/Users/SOLO/.rustup/toolchains/stable-x86_64-pc-windows-gnu/share/doc/rustfmt/                         README.md
```
So we only want an executable here, which is located in `bin`

setup_hooks.sh
Pzixel: This is pre-commit hook, it have nothing with passing build.

The idea is that guy has broken formatting, it writes `git commit` and hook just fixes all issues for him. It's better to fix an issue automatically than blame user and ask him to fix all manually.

Pzixel:
This PR is completely broken, so let's move in #55

PR #55, opened 2018-02-06
rustfmt hook

Pzixel:
Github drived crazy due to several rebases and closed my PR. So I have recreated it.


README.md
```diff
@@ -80,7 +80,7 @@ features = [ "gpu" ]
 - Create directory `C:\opencv`.
 - Copy files from `.windows` folder there.
 - Run powershell console as administrator in `c:\opencv`.
-- (***Optional, skip these steps if you don't need CUDA***) 
+- (***Optional, skip these steps if you don't need CUDA***
```
nebgnahz: Why removing half of the parenthesis?

CONTRIBUTING.md
```diff
@@ -5,5 +5,6 @@ Implement more OpenCV functions/modules and submit a PR request.
 Before submitting a PR, make sure you have formatted both C/C++ code and Rust
 code, otherwise Travis will complain.
 
-- For Rust, run `cargo fmt`
-- For C/C++, run `clang-format -i native/*`
+You have to have installed `llvm` ([download link](http://releases.llvm.org/download.html)) to make formatting happen.
```
nebgnahz: We should be concrete to mention `clang-format`, https://clang.llvm.org/docs/ClangFormat.html, instead of just giving a pointer to `llvm`. Users don't need the whole LLVM tool chain.

And I think on Mac, you can simply do `brew install clang-format`.

setup_hooks.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+rustup component add rustfmt-preview
+
+rustfmt_path=`find $HOME/.rustup/toolchains/ -wholename '*bin/rustfmt*'`
```
nebgnahz: I still do not follow why `which rustfmt` doesn't work.

setup_hooks.sh
```diff
@@ -0,0 +1,36 @@
...
+    fi
+    if [[ \"\${file}\" == *.rs ]]; then
+        rust_files+=(\"\${file}\")
+    fi
+    if [[ \"\${file}\" =~ (\.h|\.cpp|\.cc) ]]; then
+        cpp_files+=(\"\${file}\")
+    fi
+done
+echo \${rust_files[@]} | xargs --no-run-if-empty $rustfmt_path &
+echo \${cpp_files[@]} | xargs --no-run-if-empty clang-format -i &
```
nebgnahz: `--no-run-if-empty` is only available for GNU toolchain, that is, not available on Mac OS X....

README.md
Pzixel: Oops

setup_hooks.sh
Pzixel: There could be multiple `rustfmt` in system. I have 5, for example. Which one should we pick?

setup_hooks.sh
Pzixel: Do you know any alternative? I'd really live to run these tasks in parallel.

CONTRIBUTING.md
Pzixel: Sry, I wasn't aware that you shouldn't have entire toolchain on *nix. Going to fix it.

setup_hooks.sh
nebgnahz: Isn't `which rustfmt` returns the executable one that's the default (and useful) one?

```
benzh:cv-rs benzh$ which rustfmt
/Users/benzh/.cargo/bin/rustfmt
benzh:cv-rs benzh$ find $HOME/.rustup/toolchains/ -wholename '*rustfmt*'
/Users/benzh/.rustup/toolchains//nightly-x86_64-apple-darwin/bin/rustfmt
/Users/benzh/.rustup/toolchains//nightly-x86_64-apple-darwin/lib/rustlib/manifest-rustfmt-preview-x86_64-apple-darwin
/Users/benzh/.rustup/toolchains//nightly-x86_64-apple-darwin/share/doc/rustfmt
/Users/benzh/.rustup/toolchains//nightly-x86_64-apple-darwin/share/doc/rustfmt/LICENSE-APACHE
/Users/benzh/.rustup/toolchains//nightly-x86_64-apple-darwin/share/doc/rustfmt/README.md
/Users/benzh/.rustup/toolchains//nightly-x86_64-apple-darwin/share/doc/rustfmt/LICENSE-MIT
```

nebgnahz: So `bin/rustfmt` also returns two, stable and nightly, if I have two installed at the same time

setup_hooks.sh
nebgnahz: This is verbose, but helpful for dealing with OS specific issues...:

```
if [ -n "$files" ]; then
    echo $files | xargs &
fi
```

Pzixel: Yea, but how could it be piped in `&` parallel execution then?

setup_hooks.sh
Pzixel: AFAIK you can get it only if you spefically install two `rustfmt` with different default toolchain. Probably I am wrong? 

Anyhow, what's your proposal?

nebgnahz: `which rustfmt`

setup_hooks.sh
nebgnahz: if [ -n ${rust_files[@]} ]; then
     echo ${rust_files[@]} | xargs $rustfmt_path &
fi
if [ -n ${rust_files[@]} ]; then
     echo ${cpp_files[@]} | xargs clang-format -i &
fi

Pzixel: Hmm, i'm not bash pro. Does it work? This syntax seems to be broken to me.

Pzixel: Ah, I see. `&` is not a conditional operator, it just starts a new job. Ok

nebgnahz: Just updated.

Explanation: if the array has zero element, this line returns exit code 0; otherwise, because `||` is or, we run the second half, it is the same as what you have, except -P 4, meaning, at most four parallelizations.

I suggest removing `&` and `wait` to just make the format process in the foreground.

nebgnahz: Edit again: reading `xargs` manual, it seems that `-P 4` does not seem necessary.

Pzixel: See edit. `xargs` is not needed anymore

PR #56, opened 2018-02-07
Feature/flann

Pzixel:
This is a PR for FeatureMatcher, based on SIFT/SURF/...


.clang-format
```diff
@@ -38,7 +38,7 @@ BreakBeforeTernaryOperators: true
 BreakConstructorInitializersBeforeComma: false
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     80
+ColumnLimit:     120
```
Pzixel: It's because we have 120 for our rust code so we expect, that user doesn't use different resolutions for different languages :)

native/utils.h
```diff
@@ -19,4 +20,16 @@ void cv_to_ffi(const std::vector<T>& source, CVec<U>* dest) {
...
     }
 }
+
+void ffi_to_cv(const cv::Mat& source, cv::Mat* dest);
+
+template <typename T, typename U>
+void ffi_to_cv(const CVec<U*>& source, std::vector<T>* dest) {
+    dest->reserve(source.size);
+    for (size_t i = 0; i < source.size; i++) {
+        T* cell = new T();
```
Pzixel: It's probably a memory leak. Do we need `new` here? I was affraid that when T goes out of scope we can reference freed memory.

Pzixel:
> build complete a few seconds ago in 2 min 35 sec

Omg, I like what we're done with our CI. 2 minutes for build - it's amazing.

.clang-format
nebgnahz: Alright, I won't be stubborn here.

native/utils.h
nebgnahz: I think create an object `T` directly should be OK as it's pushed into the `vector` (i.e. ownership being managed by `vector`).

src/features2d/descriptor_matcher.rs
```diff
@@ -0,0 +1,104 @@
...
+        unsafe { cv_matcher_train(self.value) }
+    }
+
+    /// Returns true if there are no train descriptors
+    pub fn is_empty(&self) -> bool {
+        unsafe { cv_matcher_is_empty(self.value) }
+    }
+
+    /// Finds the best match for each descriptor from a query set
+    pub fn match_(&self, query_descriptors: &Mat) -> Vec<DMatch> {
```
nebgnahz: Curious, why is it not `match` (without underscore)?

nebgnahz:
This looks good. I need a bit more time to review the new `ffi_to_cv` and `CVecView`.

P.S. Can you break this into two PRs, one that does 120 clang-format and one that includes FLANN?

Pzixel:
Ping?

nebgnahz:
Ah, I had two comments: one about `new T`, one about `match_`. Can you reply to them first?

src/features2d/descriptor_matcher.rs
Pzixel: because it's a keyword and cannot be used as identifier. I consulted with community and they have said that it's common workaround.

Pzixel:
See edit for your second question, however I don't see anything react to in your `new T` comment.

PR #58, opened 2018-02-08
stdint

Pzixel:
I suddenly figured out that there is plenty of possible pitfalls in the code due to unknown size of builtin types. Replaces with `stdint` counterparts. I propose to look on it when someone contributes int the repo. I'l change my other PR when this one is merged.

### DIFF
--- a/native/common.h
+++ b/native/common.h
@@ -2,13 +2,12 @@
 #define CV_RS_COMMON_H
 
 #include <cstddef>
-#include <cstdint>
 #include <functional>
 #include <opencv2/core.hpp>
 
 typedef struct {
-    int32_t x;
-    int32_t y;
+    int x;
+    int y;
 } Point2i;
 
 typedef struct {
@@ -27,10 +26,10 @@ typedef struct {
 } Size2f;
 
 typedef struct {
-    int32_t x;
-    int32_t y;
-    int32_t width;
-    int32_t height;
+    int x;
+    int y;
+    int width;
+    int height;
 } Rect;
 
 typedef struct {
@@ -40,10 +39,10 @@ typedef struct {
 } RotatedRect;
 
 typedef struct {
-    int32_t v0;
-    int32_t v1;
-    int32_t v2;
-    int32_t v3;
+    int v0;
+    int v1;
+    int v2;
+    int v3;
 } Scalar;
 
 typedef struct {
--- a/native/opencv-gpu.cc
+++ b/native/opencv-gpu.cc
@@ -38,7 +38,7 @@ void* cv_gpu_hog_default() {
     return new cv::Ptr<cv::cuda::HOG>(hog);
 }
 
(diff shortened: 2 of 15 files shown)


Pzixel:
Dont't look here atm :) It's work in progress.

I'm still not sure if we should expose native `c_int` or wrap it into something else. One one hand, we should we on Rust terrain. On the other hand, we shouldn't do more than we should, in this case assume `c_int` size, and if we're wrong, system will just crack.

Pzixel:
I read [this article](https://www.reddit.com/r/rust/comments/3mc0nq/what_should_i_do_with_c_int/) and I found an interesting idea:

> I bet the library assumes int is 4 bytes anyway and isn't available on platforms where this doesn't hold. In that case there's no harm making it i32. On an exotic platform your bindings will fail to compile (if you don't use casts) but this is safer than silent data loss.

I asked if it's true for OpenCV on their site, waiting for an answer. Things probably are easier than we think they are.

nebgnahz:
Ok, let's see what assumptions OpenCV makes about `int`. Did you ask on OpenCV forum or Github Issues? The forum can be slow. I have no idea about issues.

Personally I always prefer `stdint` and making types explicit and cross-platform. That's why I started with `int32_t` in the wrapper. But here, OpenCV being the legacy code makes the decision less attractive.

I am ok with making the assertion of `c_int` the same as `i32` as your second proposal. Provide enough information so that when the assertion happens, users know what the issue is.

Pzixel:
Unfortunately, OpenCV issues are for bug reports only. I don't think I can use it to ask a question (they prohibit it in question template). You can do it on your own risk if you want so :)

I asked on the forum and on stackoverflow, however, no answers, no movements, no anything..

Pzixel:
I think we should use `c_int` until it's clear if they are supporting `i32` only. It's not our responsability to expect some size of `int`, and we shouldn't.

Pzixel:
There is no answer from officials, so I propose to just expose `c_int` as is. From SO answer:

> If the correct type is c_int, use c_int. Rust programmers shouldn't be so very fragile that they curl into the fetal position just because you use a C type in an interface to a C/C++ library.
> 
> Ask yourself this: do your users benefit from you locking out any and all potential platforms where c_int is not i32 (e.g. embedded platforms)?
> 
> If the answer is "they don't", then don't do it. If they do benefit somehow, well, you'll have to weigh that against locking out platforms yourself. Maybe the use of c_int imposes some kind of wider interface issue.

nebgnahz:
Ok, let's do this and if it causes issues (or we get new information), we can always just fix it.

Pzixel:
@nebgnahz any idea why it started failing? It says that there is some diff, but I can't get this diff locally, all files are formatted correctly...

nebgnahz:
Rust 1.24 is out, with new changes in rustfmt. You should do a `rustup` update locally.

https://blog.rust-lang.org/2018/02/15/Rust-1.24.html

Pzixel:
@nebgnahz yep, I did it, and it seems that we either have to switch to `nigthly` for a while or disable some rules.

PR #60, opened 2018-02-12
Feature/ocr

### LINKED ISSUE #59: feature request: cv::text
Specifically, I'm looking for the Tesseract support for scene text extraction.

I've been toying with a cffi binding but it's not great since it requires me to write an image to disk before I can extract text. Something I could run directly on a Mat would be ideal!

Pzixel:
Fixes #59


Pzixel:
For some reason, it works locally on fresh VM but fails on travis... Any ideas?

image

Pzixel:
@nebgnahz I need an access token in order to be able to debug build: https://docs.travis-ci.com/user/running-build-in-debug-mode/

nebgnahz:
@Pzixel Sorry for the late reply. We don't have debug build for this repository. 

```
This feature is available for private repositories and those public repositories for which the feature is enabled. To have the feature enabled for a public repository, please email us at support@travis-ci.com indicating which ones. Push access to the repository is also required.
```

I just emailed Travis for request. Stay tuned.

P.S. My solution for now is to debug using my own private repo (a different git remote).

Pzixel:
@nebgnahz please, read further :)

> To restart a job in debug mode via API, send a POST request to the job’s debug endpoint, along with your Travis CI API token in the Authorization header. As public repositories do not show the Debug button, this is the only way to restart builds in the debug mode for public repositories.

Pzixel:
Well, I managed to fix it locally. 🎉

Pzixel:
I have also added some unhandled exception handling (see `OcrHolisticWord::new`). However, I don't know hot to handle situations where TWO arguments must be retreived. I didn't find any convinient way to perform safe wrapping. Tuples are not working as expected (or I tried bad), checking only one parameter won't work because in this case some resources may leak. Well, Don't really know. If you have any idea, feel free to message me :)

nebgnahz:
Just sent you an email about the token and replies from Travis.

Pzixel:
@nebgnahz oops, I have just forgotten to remove wip status. Thank you for reminder, yep, it's ready, I guess. I moved some stuff related to `CVec` and so on in `lib.rs` for several reasons. You can freely skip these changes (I mean, pack/unpack traits and related stuff). I needed to unify them with `CResult` class, which was in another file. This leads to some additional changes that actually are not real changes.

src/core.rs
```diff
@@ -406,6 +291,15 @@ pub enum FlipCode {
 }
 
 impl Mat {
+    /// Loads `Mat` from file storage
+    pub fn from_file_storage(path: &Path, section: &str) -> Result<Mat, Error> {
+        let path = path.to_str().ok_or(CvError::InvalidPath(path.into()))?;
```
Pzixel: maybe reuse path_to_string method fom `ocr`.

src/core.rs
```diff
@@ -371,6 +254,7 @@ pub enum LineTypes {
 
 extern "C" {
     pub(crate) fn cv_mat_new() -> *mut CMat;
+    fn from_file_storage(path: *const c_char, section: *const c_char) -> *mut CMat;
```
nebgnahz: prefix with cv?

src/cuda.rs
```diff
@@ -4,6 +4,7 @@
 use super::core::*;
 use super::errors::*;
 use super::objdetect::{CSvmDetector, HogParams, ObjectDetect, SvmDetector};
+use ::*;
```
nebgnahz: Better not to import everything.

nebgnahz: Similar comments at other places with `use ::*`.

Pzixel: I think it's ok because `::*` should contain very basic features such as `CVec`, `CResult` and others. They may not be enumerated one by one, I guess.

Pzixel:
Ping?:)

Pzixel:
well, it's been 2 days without response. Could you answer if you are agree with my reasons?

Issue #20, opened 2017-11-19
No safe way to get individual elements from `Mat`


joshhansen:
Whereas the `cv::Mat` class in OpenCV provides a series of `at` methods that (if I understand correctly) return individual elements of the array, cv-rs's `Mat` provides no such methods. It seems the only way to access the actual content of the matrix is by using `data(&self)` which returns a u8 pointer to the backing data, and requires the user to understand the layout of the data, and to perform unsafe operations simply to get matrix elements. As accessing matrix elements is likely to be a common operation, it seems it would be appropriate to add equivalents to the `at` methods to cv-rs's `Mat` implementation.

nebgnahz:
You are right that the only way to access backing data within a `Mat` is to use `data(&self)`. Implementing `at` will add convenience.

With that said, I do have some concerns. OpenCV's [`at`](https://github.com/opencv/opencv/blob/2808bea7faf902e8872613cb971703dcdca4217b/modules/core/include/opencv2/core/mat.hpp#L1851) is actually complicated because of template and function overload:

```c++
    template<typename _Tp> _Tp& at(int i0=0);
    template<typename _Tp> const _Tp& at(int i0=0) const;
    template<typename _Tp> _Tp& at(int row, int col);
    template<typename _Tp> const _Tp& at(int row, int col) const;
    template<typename _Tp> _Tp& at(int i0, int i1, int i2);
    template<typename _Tp> const _Tp& at(int i0, int i1, int i2) const;
    template<typename _Tp> _Tp& at(const int* idx);
    template<typename _Tp> const _Tp& at(const int* idx) const;
    template<typename _Tp, int n> _Tp& at(const Vec<int, n>& idx);
    template<typename _Tp, int n> const _Tp& at(const Vec<int, n>& idx) const;
    template<typename _Tp> _Tp& at(Point pt);
    template<typename _Tp> const _Tp& at(Point pt) const;
```

If we are adding this API to `cv-rs`, we need to address three issues:
- [ ] What's the API of `at` in Rust binding? For example, with row and column, do we use `at(isize, isize)`, `at((isize, isize))` or `at2(isize, isize)`?
- [ ] To call C++ `at`, we need to provide the right template argument that depends on the `Mat` type, `CV_64F`, `CV_32F`, `CV_8U` will be `double`, `float`, and `uchar`, respectively.
- [ ] What's the return type? C++ returns a reference with type `_Tp`; and that need to converted to Rust types.

[...]

PR #62, opened 2018-02-19
Feature/overallfix

Pzixel:
Provide overall fixes: get rid of `num` dependency in favor of `repr(C)` enums as well as some minor fixes in FFI.

I also think we may get rid of `bytes` create. Seriosly, see this comment:

>     /// Returns individual pixel (element) information within the Mat. This
>     /// function may need type annotation to assist `FromBytes` trait.
>     ///
>     /// - If matrix is of type `CV_8U` then use `Mat.at<u8>(y,x)`.
>     /// - If matrix is of type `CV_8S` then use `Mat.at<i8>(y,x)`.
>     /// - If matrix is of type `CV_16U` then use `Mat.at<u16>(y,x)`.
>     /// - If matrix is of type `CV_16S` then use `Mat.at<i16>(y,x)`.
>     /// - If matrix is of type `CV_32S`  then use `Mat.at<i32>(y,x)`.
>     /// - If matrix is of type `CV_32F`  then use `Mat.at<f32>(y,x)`.
>     /// - If matrix is of type `CV_64F` then use `Mat.at<f64>(y,x)`.

Why not just check matrix type ourself and `match` on it? :) In this case we don't need neither `FromBytes` trait nor `bytes` dependency itself.


src/imgcodecs.rs
```diff
@@ -2,111 +2,120 @@
...
 use std::path::Path;
 use failure::Error;
 
 // =============================================================================
 //  Imgproc
 // =============================================================================
 /// ImreadModes
-#[derive(Debug, PartialEq, Clone, Copy)]
-pub enum ImreadModes {
+#[repr(C)]
```
nebgnahz: While I like the removal of `Imread` prefix in the enum fields, I have concern about this readability improvement. I think it has differed from OpenCV too much.

One way to get the best of both worlds is to improve the documentation so that users know that this ImageReadMode is ImreadModes in OpenCV. Perhaps fix the documentation with a link: https://docs.opencv.org/trunk/d4/da8/group__imgcodecs.html#ga61d9b0126a3e57d9277ac48327799c80

Pzixel: Ok. Say when you're done (or is it your only concern in this changeset?)

nebgnahz:
`Mat::at` was implemented with this issue: https://github.com/nebgnahz/cv-rs/issues/20

Since `Mat::cv_type` is only known at runtime, but `Mat::at<T>` requires a type annotation that is known at compile time, so I do not know a method to implement it with a match.

If you have a proposal, I'd love to learn about it! You have shown quite some magic in cleaning and refactoring the code base :)

src/videoio.rs
```diff
@@ -313,25 +317,26 @@ pub enum VideoWriterProperty {
...
-        ((ex & 0xFF00u32) >> 8) as u8,
-        ((ex & 0xFF0000u32) >> 16) as u8,
-        ((ex & 0xFF000000u32) >> 24) as u8,
+        (value & 0xFFu32) as u8,
+        ((value & 0xFF00u32) >> 8) as u8,
+        ((value & 0xFF0000u32) >> 16) as u8,
+        ((value & 0xFF000000u32) >> 24) as u8,
     ];
-    String::from_utf8(vec).ok()
+    String::from_utf8(vec).unwrap()
```
nebgnahz: Returns `Result<String, Error>` here?

tests/test_video.rs
```diff
@@ -0,0 +1,13 @@
+extern crate cv;
+mod utils;
+
+use cv::videoio::*;
+
+#[test]
+fn test_4cc() {
+    let code = "abcd";
```
nebgnahz: Compare some valid 4cc conversion? 

[FOURCC.org - YUV pixel formats](http://130.216.33.163/compsci773s1t/lectures/YuY2_files/intro.htm)

src/imgcodecs.rs
nebgnahz: I am done with the review.

src/videoio.rs
Pzixel: I don't believe that `[u8]` can produce invalid UTF8 string so we never gets in situation when `Err` is returned.

Pzixel:
@nebgnahz  :)

src/videoio.rs
Pzixel: I mean it's guaranteed to be valid ASCII sequence and thus valid UTF8 string

PR #65, opened 2018-02-21
Feature/global refactoring

Pzixel:
This PR doesn't contain any logic changes for easier review. I have found some places with possible memory leaks and decorated them with TODO.

It contains split of `core.rs` in several separated files (such as `mat.rs`), which is big enough, but it may be fixed with removing `frombytes` trait (could be replaced with native `at` function call), as well as breaking down of `opencv-wrapper` in multiple little files. Looks much nicer to me:

image

There is also `gpu` -> `cuda` rename to be more spesific as well as because OpenCV may also provide support for OpenCL, TensorFlow or something else in future.


src/highgui.rs
```diff
@@ -114,3 +117,20 @@ pub enum MouseEventType {
     /// Positive/negative means right and left scrolling.
     MouseHWheel = 11,
 }
+
+/// Provides some highgui functionallity
+pub trait Shower {
```
Pzixel: Is not the best name

nebgnahz: Verb may be better, like `Show`.
Or just `HighGui` or `Gui`?

native/cuda/main.cc
```diff
@@ -1,10 +1,9 @@
 #include <opencv2/cudaobjdetect.hpp>
 
-#include "opencv-gpu.h"
-#include "opencv-wrapper.h"
-#include "utils.h"
+#include "../utils.h"
+#include "main.h"
```
nebgnahz: main suggests that there might be a main function and this is a binary executable.

we could do `native/cuda/cuda.[h|cc]`

native/cuda/main.h
```diff
@@ -68,7 +65,6 @@ int cv_gpu_cascade_get_min_neighbors(cv::Ptr<cv::cuda::CascadeClassifier>*);
 Size2i cv_gpu_cascade_get_max_object_size(cv::Ptr<cv::cuda::CascadeClassifier>*);
 Size2i cv_gpu_cascade_get_min_object_size(cv::Ptr<cv::cuda::CascadeClassifier>*);
```
nebgnahz: Mmm, since we are renaming everything to cuda, what about changing these to `cv_cuda_*`?

nebgnahz:
I mainly have some naming suggestions.

PR #66, opened 2018-02-25
WIP: Feature/submodules

Pzixel:
@nebgnahz 
This is a feature where I'm trying to exclude user from configuration as much as possible.

In this change, following things are done:

1. Instead of cloning repo in build scripts `opencv` and `opencv_contrib` are included as git submodules. That allows us to reference precise hash commit, which is clear for anyone watch repo. 
1. Thus, we have all knowledge about `opencv` location and therefore we can avoid asking user for environment configuration, such as `OPENCV_DIR` and `OPENCV_LIB` variables.
1. Thus, we can have several versions of OpenCV compiled separately. For example, one for `cuda` and one without it. ATM, it's not possible, for several reasons.
1. Some other benefits such as better control over build process, e.g. we can check that if `CUDA` switch is enabled then CUDA opencv is compiled. Before this changes we just was hoping that user passed right `-EnableCuda` parameter to script.

As a bonus, we have easier Linux configuration, because build process is unified across platforms, instead of windows-specific `ps1` and linux `sh` scripts.

However, I need you help. I'm stuck because I can't make it work without copying files to `/usr/local/include/`. One of main points is that `build.rs` cannot run `sudo` commands. 

We need to invent something to make build happen on Linux. It actually build right now, but it fails at runtime when trying to actually execute dynamic library code.

Could you please fix this issues? When it's done I fix some issues such as scripts, docs and better naming/etc and provide resulting PR. But atm i'm really stuck with ubuntu behaviour.


nebgnahz:
Without copying, it might have to do with setting `$LD_LIBRARY_PATH` or other tricks to let the OS knows where to find the library. I am outside now and will take a look ASAP.

Pzixel:
Yep, I found that it's the most probable cause.
Here are some thoughts: https://stackoverflow.com/questions/17689528/update-ldconfig-cache-without-root-permission

The best way is probably ask user to run `ldconfig` once first build is complete.

nebgnahz:
Hey @Pzixel, thanks for the efforts for making the build process painless. Before we change too far, can you clarify a bit about the future process of using `cv-rs`?

First, one includes `cv = { git = "https://github.com/nebgnahz/cv-rs.git" }`. When building the project, it will clone all opencv submodules and build the library (`opencv_world` for all systems).  So from the user perspective, while it took a while to build the whole project, the user does not need to do any extra configurations (perhaps except `ldconfig`).

I have several concerns about this direction:

1. If I have several CV applications, each will need to download and build opencv.
2. This conflicts with package management systems. On Mac OS X, many users, me included, simply do `brew install opencv3 --with-contrib --with-cuda`. So installing OpenCV isn't a big issue; and brew is a much better package management systems than our ad-hoc one. Similar argument goes to `apt` on Linux and `choco` for Windows. I understand there are limitations with these package management systems, such as `choco` does not support `cuda` and `opencv_contrib` well. But many users have already been familiar with these package manager. This does require a proper configuration of `OPENCV_INCLUDE` and `OPENCV_LIB`.
3. I think when you compile a `Release` version of CV, it will re-compile OpenCV again.

Now replying your proposals:

> Instead of cloning repo in build scripts opencv and opencv_contrib are included as git submodules. That allows us to reference precise hash commit, which is clear for anyone watch repo.
> Thus, we have all knowledge about opencv location and therefore we can avoid asking user for environment configuration, such as OPENCV_DIR and OPENCV_LIB variables.

I like precise hash commit for deterministic behaviors.

> Thus, we can have several versions of OpenCV compiled separately. For example, one for cuda and one without it. ATM, it's not possible, for several reasons.

This is possible at this moment. You can change the system installed OpenCV libraries (either symlinks or copy).

> Some other benefits such as better control over build process, e.g. we can check that if CUDA switch is enabled then CUDA opencv is compiled. Before this changes we just was hoping that user passed right -EnableCuda parameter to script.

Fair.

> As a bonus, we have easier Linux configuration, because build process is unified across platforms, instead of windows-specific ps1 and linux sh scripts.

I failed to see the argument against a "complex" Linux configuration. Yes, one needs to build from source, but that's quite common for Linux users. Also there is custom apt sources, and Linux brew to rescue.

> Could you please fix this issues? When it's done I fix some issues such as scripts, docs and better naming/etc and provide resulting PR. But atm i'm really stuck with ubuntu behaviour.

I am trying to debug this on my Linux server, and am still waiting for `Compiling cv v0.2.1 (file:///home/benzh/repos/cv-rs)` (it has been ~30 minutes).

nebgnahz:
Update: I was able to run binary on my Linux server. I did two changes:

```Rust
let opencv_world_without_extension = "opencv_world"; // not libopencv_world
```

```bash
cv-rs $ LD_LIBRARY_PATH=artifacts/default/lib/ cargo test --test  test_basic_ops
```

Pzixel:
Answering your concerns.

> First, one includes cv = { git = "https://github.com/nebgnahz/cv-rs.git" }. When building the project, it will clone all opencv submodules and build the library (opencv_world for all systems). So from the user perspective, while it took a while to build the whole project, the user does not need to do any extra configurations (perhaps except ldconfig).

As far as I see build time isn't increased because of building `world`, but it's profitable. We can use CI build time as reference.

> If I have several CV applications, each will need to download and build opencv.

> This conflicts with package management systems. On Mac OS X, many users, me included, simply do brew install opencv3 --with-contrib --with-cuda. So installing OpenCV isn't a big issue; and brew is a much better package management systems than our ad-hoc one. Similar argument goes to apt on Linux and choco for Windows. I understand there are limitations with these package management systems, such as choco does not support cuda and opencv_contrib well. But many users have already been familiar with these package manager. This does require a proper configuration of OPENCV_INCLUDE and OPENCV_LIB.

I think, the main flow is one `cv-rs` that is used for all other applications. However, it's always possible to add feature that skips installation and hopes that it's installed somewhere. It's not hard to implement, we always can skip intallation step if user want so.

> I think when you compile a Release version of CV, it will re-compile OpenCV again.

It's already building in release mode because we aren't supposed to debug OpenCV. If user want so, it may always manually change `target`  in `build.rs` script.

> This is possible at this moment. You can change the system installed OpenCV libraries (either symlinks or copy).

It's not possible on Windows because you may have only one OPENCV_LIB at every moment.

> I failed to see the argument against a "complex" Linux configuration. Yes, one needs to build from source, but that's quite common for Linux users. Also there is custom apt sources, and Linux brew to rescue.

I don't say it's complex, but it differs from Windows one. I think there should be one single source of truth. Different scripts gives different flows. In build script we minimize it - all main steps are common, there is only some difference in `post_build`, which is fine. It's much better than completely unerlated `install.sh` and `install.ps1`.

> I am trying to debug this on my Linux server, and am still waiting for Compiling cv v0.2.1 (file:///home/benzh/repos/cv-rs) (it has been ~30 minutes).

just use `cargo build -vv`. For some reason cargo doesn't show `cc` output until verbosity is set.

---

The main motivation of this change is allow users simply write `git clone && cd cv-rs && cargo build -vv` and be able to just drink tea while environment gets configured automatically. It's the main flow and it should be the simpler possible. And it's doesn't force us to recompile `opencv` every time something changes. But it gives at least three benefits: 

- We always are set and newcomers are easy to use this lib, because they just should run one singleliner above
- We are always building against known opencv version. Does it work with opencv 3.0.0? Will it work with 3.5.0? I don't know. We can be sure about opencv version only if we compile it ourself.
- User may want to use different opencv versions. In this case we just have our pocket opencv while system can use completely unrelated one.

My point is make common things simple, and complex one possible. I think this PR is the right step in this direction. We can replace all out instllation section in readme via *write this singleliner and just get some coffee" for most users. For ones who want to share OpenCV installation with some other software it's always possible to skip this installation step. It's dangerous and may cause runtime errors, but possible. However, I'd like to wait until someone real appear and say "guys, I have this need right now", because I think it may be not demanded at all.

The next step in my plans was `bindgen` opencv C API and provide bindings to entire OpenCV (or most of it). It would allow us to lower burden of C++ code support as well as faster feature introduction.

nebgnahz:
### Experience so far

I am currently playing with this change and see how it feels. The one-liner is great for a new user to start a new project. But when it comes to run it, I have to fix the issue with load path.

For a toy project that uses `cv-rs` as a dependency, I have to modify `LD_LIBRARY_PATH` as follows:

```
LD_LIBRARY_PATH=/home/benzh/.cargo/git/checkouts/cv-rs-2ef8cae194ab26b5/c44ef99/artifacts/default/lib/ cargo run
```

Other than printing this directory out when executing `build.rs`, I am not sure how else we can get the path.

---
### About Release vs Debug:

> It's already building in release mode because we aren't supposed to debug OpenCV. If user want so, it may always manually change target in build.rs script.

Well, for someone using `cv-rs` as a dependency, if the project itself is in `debug`, all dependencies are built in `debug` mode. For a `release` build, that would be a different build. Actually this is not a big issue because OpenCV library will built into the same folder (`.cargo/git/checkouts/cv-rs-2ef8cae194ab26b5/c44ef99/artifacts/default/`). So the build is fast because the library has already been built.

---
### cv-rs updates

Whenever `cv-rs` has an update, even as small as a [documentation change](https://github.com/nebgnahz/cv-rs/commit/e8d8d10b74d079d11fa41b348c2fa9992f89209c), if the user of this library do a `cargo update` and then `cargo build`, because the artifacts folder is now changed, it leads to another compilation and installation of OpenCV (hence another cup of tea/coffee).

Before: `/home/benzh/.cargo/git/checkouts/cv-rs-2ef8cae194ab26b5/c44ef99/artifacts/default/`
After: `/home/benzh/.cargo/git/checkouts/cv-rs-2ef8cae194ab26b5/e8d8d10/artifacts/default`

---
### Design
> Simple things should be simple, complex things should be possible 

Ah ha, I like the Alan Kay quote but have a different interpretation. To me, the simple thing is a simple OpenCV wrapper that has little magic. But I see your point of simple being no fuss with OpenCV configurations. In the past, I have also struggled with installing OpenCV; and I know a lot of others are probably still struggling.

But still, another Unix philosophy I like is **do one thing and do it well**. So this `cv-rs` wrapper should probably not do what a package manager is doing.

By the way, do you mind if I (or you can) post this design and integration on user.rust-lang? We could learn from others how they handle the external dependencies when writing Rust FFI. If you feel that this crate is still in an early stage and don't want to get too much attention (like I was having such feeling before), I am OK with waiting a bit longer. We can work out this issue ourselves.

Pzixel:
@nebgnahz I'l answer from the bottom to top.

> By the way, do you mind if I (or you can) post this design and integration on user.rust-lang? We could learn from others how they handle the external dependencies when writing Rust FFI. If you feel that this crate is still in an early stage and don't want to get too much attention (like I was having such feeling before), I am OK with waiting a bit longer. We can work out this issue ourselves.

The main reason beside create a great library is learn to write complex rust programs. Of course, you can post it, I'd love to learn from it.

> Whenever cv-rs has an update, even as small as a documentation change, if the user of this library do a cargo update and then cargo build, because the artifacts folder is now changed, it leads to another compilation and installation of OpenCV (hence another cup of tea/coffee).

Hmm, this guy really kills the whole idea. I think the post on `rust-lang` could enlighten better design. Rebuilding after update is definitely  not what I'd like to have.

> For a toy project that uses cv-rs as a dependency, I have to modify LD_LIBRARY_PATH as follows:


That's also is a no-way for the project.

Without answering these considerations this issue cannot move. So I'd appreciate if you create a post with the question. Thank you for revealing these flaws, we will either solve them or completely deprecate this approach.

However, even if this idea is completely wrong, I think `git submodules` theme would be great anyway, because modules won't take any space until user `git init` them manually, but eliminates "download opencv from somewhere" step.

Pzixel:
After some thinking, I have following proposal:

1. Use git submodules to get sure we are on the right hash. It also simplifies lots of things, for example we don't need any environment variables anymore
1. Do not move opencv build in build.rs
1. However, get rid of `sh` and `ps` scripts. How? I think it could be done via custom `cmake` file. We just move all build steps in `cmake` file which simplifies CI as well as README documentation.
1. I might forget something, may add more later.

As result: git submodules allow us to reference its exact version, but doesn't force user to use them. Build script will use OPENCV_LIB that is referencing target installation folder. It either is set by build script or by user manually. In this case user have a choice of using some shared OpenCV by setting this variable or allow it to be set by the script. 

Finally, I still think that we should either build `opencv_world` on both platforms or on neither one.

P.S. However, it's only my thoughts and I still waiting for community opinion that you wanted to get on `rust-lang`. Thank you for writing such an exhaustive reply, I appreciate it.

nebgnahz:
Hey, are you currently blocked on this PR? I mean, would your bindgen would depend on the merge of the changeset here? If so, maybe we can break this into a couple smaller ones and get them going, such as use git submodules for opencv code.

Sorry that I have other commitments coming up this week and couldn't finish the post to user.rust-lang.

Pzixel:
Hey :) 

I'm soft-blocked i.e. I'd like to have some changes in master, however, it's not required. It's just like "I don't want to resolve these conflicts", it's possible, but noone likes it :)

However, I have nothing to do, this is why it seems that I can give it a shot.

I think this PR should be closed anyway, because one one hand these changes are no-no in upstream (as we decided), but on the other hand some of these changes could be useful in future, so I don't want to squash them out. So this PR should be closed with all changes as they are in case they would required somewhere later. The question is what should go in this new PR, beside git modules.

Pzixel:
Bad news come:

> OpenCV C API is dead since 2010. please do not attempt to write any new code this way

> (write your own wrappers, yes)

> ((and even then, you'll have to use a c++ compiler, exceptions and whatnot))

Thus we can't use bindgen, because it doesn't work ideally with pure C functions, but it's extremely bad in working with C++ code.

However, I'm still wonderng if moving all natives in separated crate worth it.

Pzixel:
Then, will try to use [cpp generator](https://github.com/rust-qt/cpp_to_rust/tree/master/cpp_to_rust/cpp_to_rust_generator)

nebgnahz:
Yes, OpenCV official guides discourage developing and using C APIs. But with Rust FFI, this is by far the only way to get it working. Rust bindgen actually works with C++ project (to a limited extent). My previous attempt of autogenerating bindings (roughly 1.5 yrs ago) were not that successful. Maybe now Rust side tools are more mature.

Pzixel:
@nebgnahz I don't believe bare metal bindgen can generate C++ bindings. They are trying, but not yet. I'm pretty sure that this project that I mentioned is built on top of bindgen itself. 

> OpenCV official guides discourage developing and using C APIs

I know it, but I had a bit different information that it's not the most recommended way, but it's viable, especially when I found information that they are going to write a C API generator, that automatically generates FFI compatible API from C++ code.

> Let me also comment, since I'm the author of that provocative statement =)
> 
> Shervin is right, the C API is not developed for a long time. All the new stuff has the C++ API, and it is not backported to the C. So, C API becomes obsolete and causes pain in the neck, since it should be maintained. Thus, the decision is made to mark it as deprecated. But let me clarify the point:
> 
> Actually C API will stay in the library, but it will not be available by default. It will be put into separate header files, that should be manually included by user. So, it would be possible to compile the legacy code, but you should add these compatibility includes manually.
> It is possible that OpenCV will generate compiler warnings in case you're using C API. These warnings will notify you that you're using deprecated API, and you should migrate to C++.
> At the same time in the long term C API may return to the OpenCV, so it can be used from the C code (e.g. low-level or embedded programming). But the plan is to autogenerate C wrappers for C++ code, as we do for Java and Python. This way C API will be always up-to-date, since it is generated from the C++ headers automatically. But this is a long-term dream, not for OpenCV 3.0, and if you want to work on that, please let us know (http://opencv.org/contribute.html).
> 
> To summarize, 3.0 will not bring dramatical public API changes. It will simply implement the 2.x design concepts to the full degree. Our goal is to make the transition as smooth as it is possible. You will need to change a couple of includes, may be update some calls. But we'll use this possibility (major release) to break the compatibility in order to improve the structure and make the API more consistent.

Thus, I didn't expect it's *that* bad. But it's ok, let's try a different tool.

PR #67, opened 2018-03-01
Feature/submodules working

Pzixel:
This is an attempt of clean fresh start of `submodules` idea.

However, I didn't succeed in moving from shell scripts to CMake. Here is my latest (and incomplete) attempt:

```
cmake_minimum_required(VERSION 2.8)
set_directory_properties(PROPERTIES EP_BASE ${GLOBAL_OUTPUT_PATH})
project(opencv_native)

include(ExternalProject)


# Sets global output directory for single configuration (GCC)
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${GLOBAL_OUTPUT_PATH})
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${GLOBAL_OUTPUT_PATH})
set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${GLOBAL_OUTPUT_PATH})

# Sets global output directory for sub-configurations (msvc, mingw)
foreach(OUTPUTCONFIG ${CMAKE_CONFIGURATION_TYPES})
    string(TOUPPER ${OUTPUTCONFIG} OUTPUTCONFIG)
    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${GLOBAL_OUTPUT_PATH})
    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${GLOBAL_OUTPUT_PATH})
    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${GLOBAL_OUTPUT_PATH})
endforeach(OUTPUTCONFIG CMAKE_CONFIGURATION_TYPES)

#adding openv
ExternalProject_Add(opencv
	DOWNLOAD_COMMAND ""
	SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../opencv
	CMAKE_ARGS -DWITH_CUDA=OFF -DCUDA_ARCH_BIN=5.2 -DCUDA_ARCH_PTX= -DBUILD_opencv_java=OFF -DBUILD_opencv_python=OFF -DBUILD_opencv_python2=OFF -DBUILD_opencv_python3=OFF -DBUILD_TESTS=OFF -DBUILD_PERF_TESTS=OFF -DBUILD_DOCS=OFF -DBUILD_EXAMPLES=OFF -DINSTALL_CREATE_DISTRIB=ON
)
```

However, it doesn't start `opencv` build. I'm not familiar with cmake and though I googled a lot I didn't be able to make it work. Maybe you can do better.

### DIFF
--- a/.ci/travis_build_opencv.sh
+++ b/.ci/travis_build_opencv.sh
@@ -12,14 +12,7 @@ INSTALL_PREFIX=$HOME/usr
 if [[ ! -e $INSTALL_FLAG ]]; then
     TMP=$(mktemp -d)
     if [[ ! -d $OPENCV_BUILD ]]; then
-        curl -sL ${URL}  > ${TMP}/opencv.zip
-        unzip -q ${TMP}/opencv.zip
-        rm ${TMP}/opencv.zip
-
-        curl -sL ${URL_CONTRIB}  > ${TMP}/opencv_contrib.zip
-        unzip -q ${TMP}/opencv_contrib.zip
-        rm ${TMP}/opencv_contrib.zip
-
+		git submodule update --init --recursive
         mkdir $OPENCV_BUILD
     fi
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "opencv_contrib"]
+	path = opencv_contrib
+	url = https://github.com/opencv/opencv_contrib.git
+[submodule "opencv"]
+	path = opencv
+	url = https://github.com/opencv/opencv.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,8 @@
 [submodule "opencv_contrib"]
 	path = opencv_contrib
 	url = https://github.com/opencv/opencv_contrib.git
+	shallow = true
 [submodule "opencv"]
 	path = opencv
 	url = https://github.com/opencv/opencv.git
+	shallow = true
--- a/.windows/mingw_build_OCV.ps1
+++ b/.windows/mingw_build_OCV.ps1
@@ -56,11 +56,7 @@ Write-Host "INSTALL OPENCV AT $OPENCV_DIR"
 mkdir $OPENCV_BUILD_DIR -ErrorAction SilentlyContinue
 mkdir $OPENCV_DIR -ErrorAction SilentlyContinue
 
-$oldErrorAction = $ErrorActionPreference
-$ErrorActionPreference = "SilentlyContinue"
-git clone -b $OPENCV_VERSION_TAG --depth 1 https://github.com/opencv/opencv.git
-git clone -b $OPENCV_VERSION_TAG --depth 1 https://github.com/opencv/opencv_contrib.git
-$ErrorActionPreference = $oldErrorAction
+git submodule update --init --recursive
 
 Push-Location -Path $OPENCV_BUILD_DIR
 $CMakeArgs = $CMAKE_OPTIONS + ("-DCMAKE_INSTALL_PREFIX=$OPENCV_DIR", "-DCMAKE_BUILD_TYPE=Release", "-DOPENCV_EXTRA_MODULES_PATH=$pwd\opencv_contrib\modules", $REPO_LOCATION)
--- a/.windows/msvc_2_build_OCV.ps1
+++ b/.windows/msvc_2_build_OCV.ps1
@@ -1,83 +1,68 @@
 param([Parameter(mandatory=$true)][bool] $EnableCuda, [Parameter(mandatory=$true)][string] $Compiler)
-$CudaSwitch = If ($EnableCuda) {"ON"} Else {"OFF"}
+$CudaSwitch = If ($EnableCuda) { ON } Else { OFF }
 #SCRIPT CONSTANTS
 $pwd = Get-Location
-$REPO_LOCATION = "$pwd\opencv"
-$OPENCV_VERSION_TAG = "3.4.0"
+$REPO_LOCATION =  $pwd\opencv 
+$OPENCV_VERSION_TAG =  3.4.0 
 $CMAKE_CONFIG_GENERATOR;
 
-if ($Compiler -eq "vc14") {
-    $CMAKE_CONFIG_GENERATOR = "Visual Studio 14 2015 Win64"
+if ($Compiler -eq  vc14 ) {
+    $CMAKE_CONFIG_GENERATOR =  Visual Studio 14 2015 Win64 
 }
 else {
-    if ($Compiler -eq "vc15"){
+    if ($Compiler -eq  vc15 ){
         if ($EnableCuda) {
-            throw "Cuda with VS2017 is not supported"
+            throw  Cuda with VS2017 is not supported 
         }
-        $CMAKE_CONFIG_GENERATOR = "Visual Studio 15 2017 Win64"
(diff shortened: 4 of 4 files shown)


nebgnahz:
The gitmodules looks good.

But I am confused about the CMake: we are creating a new CMake to just wrap OpenCV's CMake?

Pzixel:
Yep, that was the original idea, but it seems to be not working for some reason. Maybe I'm just not so good in making make scripts (it's my first attempt ever), but it seemed to be the only cross-platform way to configure make. I just don't like we have 3 scripts that we have to maintain synchronized, and changes in one script may be forgotten to be added in anothers, common copypaste problem, you know :)

nebgnahz:
Sure, having a cross-platform build system is nice and CMake does a reasonable job.

But As I am looking at the scripts we have for now, there are two scripts across different platforms: one for installing CUDA, one for building OpenCV (mostly just calling OpenCV's CMake). The benefit of having this meta-CMake seems to be wrapping OpenCV CMake command line argument in a cross-platform way. Given that is only a couple of lines of code, I do not know how beneficial it is to work on this. And copy-paste problems should be detected by CI.

Again, for me, I would also like a cross-platform build process because I do not work on Windows, but this library should support Windows users (btw, thank you for all the Windows build improvements), but I felt like working on CMake is not a good investment of time. P.S. I have worked for [a project](https://github.com/damellis/ESP/blob/master/CMakeLists.txt) using CMake for cross-platform build, CMake is not friendly to work with :)

Pzixel:
@nebgnahz yeah, I see :)

> Given that is only a couple of lines of code, I do not know how beneficial it is to work on this. And copy-paste problems should be detected by CI.

But we still make mistakes with these couples of lines. The most notable example, I guess, is compiling world in one case and not compiling it in another one. The most consistent way is use it everywhere or nowhere.

However, I agree that we probably have higher priorities. 

Another question is about codegen. [See my question on `opencv.org` ](http://answers.opencv.org/question/185693/is-c-api-alive-or-not/). Main OpenCV developer noted that without codegen this initiative may not quit "toy example" state. And it's kind an issue. We may automate a lot of things, such as:

1. All primitive types, such as CvRect, CvPoint, ...
1. Constructors and destructors for all types. Automatic `drop` implementation.
1. Right signatures for functions as well as comments for functions and parameters.

This is the main problem I see right now, it's quite strategic.

Pzixel:
Ping?:)

We also might look on [helix](http://blog.skylight.io/introducing-helix/#backtofast_blank) and implement some codegen via macros. 

@berak is right, without some sort of codegen it would stay at "toy project" state. If we want some acceptance we have to solve all the complexity around the library.

nebgnahz:
@Pzixel I am all in for a bindgen solution (either using Rust bindgen or adapting existing solutions for Python/Java/Go). I fully agree that without doing that, the manual solution is not scalable for other modules and this will stay as a toy project.

There was another project doing the "adapt Python build process to Rust", https://github.com/kali/opencv-rust/. Last time I checked, it worked better for OpenCV 2.x, not so good for 3.x. For my use case, I need 3.x. And because I wasn't targeting at all OpenCV APIs but only a few modules that I need, I started this project and focused on a simple wrapper that does not do magic bindgen at all. You can take a look at that project and see if the approach is viable.

P.S. I am already over committed recently. Apologies. While I am happy to review your changes and having discussions, implementation-wise, I would not be able to do much (I will be much much much more free after May...)

Pzixel:
Yep, I have seen this project, however it looked abandoned and I didn't get how it works. Now I know how bindgen works and I understand this project, but it's a no way for us - I've talked with many people that are using bindgen and they think that bindgen should be user as CLI tool only, no `build.rs` with bindgen ever.

> There is another project doing the "adapt Python build process to Rust", https://github.com/kali/opencv-rust/. Last time I checked, it worked better for OpenCV 2.x, not so good for 3.x. For my use case, I need 3.x. And because I wasn't targeting at all OpenCV APIs but only a few modules that I need, I started this project and focused on a simple wrapper that does not do magic bindgen at all. You can take a look at that project and see if the approach is viable.

So am I, I needed some features to make image database, but involved in library development a bit :)

However, I think my work with CVec, CResult and other types could be reused for this codegen, so they won't be useless waste of time.

We may adopt some code from `opencv-rs` project to make it compile with `3.X` and then introduce more modules. Or should we?

The main reason for codegen is that we shouldn't write our code. But Rust has completely different approaches as C++ code. You cannot generate ideomatic Rust code with bindgen, it's only possible to do manually, e.g. remember all these crazy enums (thay look much better in our wrapper) as well as class hierarchy that looks like trait objects and independed structs (see matchers). It defenitely may be generated, but I don't believe it may be done by bindgen heuristics. 

I'm kinda lost here. We can't provide the whole opencv wrapper written manually but we cannot generate ideomatic wrapper. All in all, we **probably** may enhance bindgen heuristics to make it work with hierarchies, and that **probably** would allow us to generate wrappers. But again, who knows what else must be enhanced to make it work in a nice way and how much effort should we spend.

Sorry for chaotic writing, I was just writing my thoughts as they come.  The general idea is that we probably can spend several months or even years in enhancing bindgen which probably will be accepted in upstream and which probably will solve some of problems. And this conclusion makes me sad: it shows that it's probably impossible to do by just us two.

Pzixel:
However, I'd like to end writing on something positive:

https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/hal/interface.h#L21-L31

OpenCV actually works only on `int=i32` platforms so we can easily remove all our `c_int`'s from API.

berak:
@Pzixel , i somehow got pinged here, let me add some comments:
* no i'm not a "main opencv developer", i'm just loud on answers.opencv.org
* "We may automate a lot of things, such as: All primitive types, such as CvRect, CvPoint, ..." 
 i'd try the opposite way: map fundamental types manually; once you got that, it should be a breeze, generating functions like `add(src1, src2, result)` (and again, this should be based on c++ types, not the arcane c-api)
* i tried to eat my own dog-food recently, trying to automate opencv wrappers for [gocv](https://github.com/hybridgroup/gocv/), but after putting in like 60 hrs on some (current 1038 lines) python script, based on gen2.py, results are only 70%correct. this IS terribly hard ! 
whatever codegen you use, in the end it's a collection of "edge cases", that noone but the author (and you !) may have encountered. 
* a single person may try a pr with a few wrappers, and this will need only like overview of maybe 20% of the complexity, but a full-blown codegen needs 101% ;( 
(on the other hand github interaction with a 1-function pr WILL tire out any maintainer, also the sun will melt, before there's any real progress, that way ...)
* i somewhat lost contact to the rust-sphere (shame on me), but it seems it needs 2 layers of wrapping: c++ args to c, -> c args -> rust, and then the reverse for any results: rust ->  c,  c -> c++ 
<br>

**in the end: do things in small steps. do try with codegen !** 
(of your favourite sort), but **be sceptical**. 

feed it with small input (a few functions (classes, if you dare even !) at a time, check the output, do tests.)

i still think, playing with some kind of codegen is a good idea. even if it may not be **immediately** productive, (aka: merged pr's until you reach the 101% stage) , 

for sure, it will make you understand both rust AND opencv concepts better (even if that's bought terribly expensive ;)

PR #72, opened 2018-05-01
extract nonfree dependencies into a feature

peahonen:
to make build succeed on Ubuntu by default


Pzixel:
There is no such concept as `nonfree`. You can try `cfg` under `contrib` flag: https://github.com/opencv/opencv_contrib.

I believe all of them all nonfree.

If they don't match, then ok, but still need some better naming, I believe.

Cargo.toml
```diff
@@ -24,3 +24,4 @@ gcc = "0"
 [features]
 cuda = []
 tesseract = []
+nonfree = []
```
Pzixel: There is no such concept as nonfree. You can try cfg under contrib flag: https://github.com/opencv/opencv_contrib.

I believe all of them all nonfree.

If they don't match, then ok, but still need some better naming, I believe.

tests/test_features2d.rs
```diff
@@ -5,6 +5,7 @@ use cv::*;
 use cv::features2d::*;
 use utils::*;
 
+#[cfg(feature = "nonfree")]
 #[test]
 fn mser_lenna_detect() {
```
Pzixel: mser is not `nonfree` feature.

tests/test_features2d.rs
```diff
@@ -38,6 +41,7 @@ fn sift_lenna_detect_and_compute() {
     assert_eq!(keypoints.len() as i32, descriptors.rows);
 }
 
+#[cfg(feature = "nonfree")]
```
Pzixel: flann is not nonfree

tests/test_text.rs
```diff
@@ -50,6 +50,7 @@ mod tesseract {
     }
 }
 
+#[cfg(feature = "nonfree")]
```
Pzixel: hmm is not nonfree

tests/test_text.rs
```diff
@@ -77,6 +78,7 @@ fn ocr_hmm_test() {
     assert_ne!(reslen, 0); // do not check actual recognized text, waiting for fix: https://github.com/opencv/opencv_contrib/issues/1557
 }
 
+#[cfg(feature = "nonfree")]
```
Pzixel: holistic is not nonfree

Cargo.toml
peahonen: Sure, a better name would be appropriate.
Original problem was the fact that Ubuntu package of OpenCV seems to leave out some of the modules used by this package.
I just assumed they were not free enough for Debian.

I'm new to this, and would be happy with any other solution as well to make this build on Ubuntu.

There seems to be some interdependencies with tesseract to be resolved also.

Pzixel: Well, [this script](https://github.com/nebgnahz/cv-rs/blob/master/.ci/travis_build_opencv.sh) works on travis which is using Ubuntu image. So if it's your problem then you can try to adopt it for you.

However, better separation could be a good thing. Just some thoughts.

peahonen: Thanks for the link. I see that builds OpenCV from source.
While that would enable the complete functionality and be a viable option on Ubuntu for some, I would need to be able to use the Ubuntu packaged version. This calls for a mechanism to disable some of the dependencies in the wrapper package. At least optionally, if not by default.

Is this general approach of feature configuration the right way to do this? If only the name(s) were more appropriate?
Where is the current set of feature names derived from? It's not 1:1 OpenCV module names either.

Pzixel: @peahonen no, it's not only naming thing.

For example, I don't see why `mser` should be cutted out with cfg. There is no reason why it couldn't work on ubuntu. Can you confirm that when you download prebuilt package it doesn't contain MSER implementation?

mwillsey:
Thanks @peahonen for making it build on debian; that was super useful for me! If there's any other work that's needed on this PR, I'd be happy to chip in.

PR #73, opened 2018-05-01
Add GaussianBlur, erode, dilate and threshold

peahonen:


src/imgproc.rs
```diff
@@ -89,6 +116,22 @@ pub enum HistogramComparisionMethod {
     KullbackLeiblerDivergence = 5,
 }
 
+/// ThresholdTypes used in
+/// [threshold](../struct.Mat.html#method.threshold).
+#[repr(C)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+#[allow(non_camel_case_types, missing_docs)]
+pub enum ThresholdType {
+    THRESH_BINARY = 0, // \f[\texttt{dst} (x,y) =  \fork{\texttt{maxval}}{if \(\texttt{src}(x,y) > \texttt{thresh}\)}{0}{otherwise}\f]
```
Pzixel: Please, provide more "rustish" names (see others enums as example).

If these comments are valuable, please make them doc comments that they appear in documentation.

src/imgproc.rs
```diff
@@ -89,6 +116,22 @@ pub enum HistogramComparisionMethod {
...
+    THRESH_BINARY_INV = 1, // \f[\texttt{dst} (x,y) =  \fork{0}{if \(\texttt{src}(x,y) > \texttt{thresh}\)}{\texttt{maxval}}{otherwise}\f]
+    THRESH_TRUNC = 2, // \f[\texttt{dst} (x,y) =  \fork{\texttt{threshold}}{if \(\texttt{src}(x,y) > \texttt{thresh}\)}{\texttt{src}(x,y)}{otherwise}\f]
+    THRESH_TOZERO = 3, // \f[\texttt{dst} (x,y) =  \fork{\texttt{src}(x,y)}{if \(\texttt{src}(x,y) > \texttt{thresh}\)}{0}{otherwise}\f]
+    THRESH_TOZERO_INV = 4, // \f[\texttt{dst} (x,y) =  \fork{0}{if \(\texttt{src}(x,y) > \texttt{thresh}\)}{\texttt{src}(x,y)}{otherwise}\f]
+    THRESH_MASK = 7,
```
Pzixel: If you are going to comment enum please add a comment for this member too.

src/imgproc.rs
```diff
@@ -355,6 +398,73 @@ impl Mat {
         Mat::from_raw(m)
     }
 
+    /// Threshold
+    ///
+    /// TODO: https://docs.opencv.org/3.4.0/d7/d1b/group__imgproc__misc.html#gaa9e58d2860d4afa658ef70a9b1115576
+    pub fn threshold(&self, thresh: f64, maxval: f64, ttype: ThresholdType) -> Mat {
```
Pzixel: `treshold_type: ThresholdType`

src/imgproc.rs
peahonen: They were from the original OpenCV definitions. Actually there's no point in duplicating them here.

Pzixel: Our overall intent was providing greater user expirience. If this is public enum as is intended to be used from rust code then it should use rust enum guidelines. This is why all public enums works this way. Please, use existing project code style.

Pzixel:
For some reason I wasn't mentioned that you have fixed the code. Sorry for late merge. Great job.

Issue #74, opened 2018-05-19
Is it viable to make the inner field in the Mat structure public?


AgustinCB:
I understand why it's restricted to the crate and it makes sense.

However, I'm writing an integration in rust with the dlib library and it'll be very handy if that field were public. It will make much easier the connection between dlib and opencv. Another possibility would be to have a method that returns it, possibly marked as unsafe.

That being said, again, I understand why that's not desirable. And in such a case, I'll just fork the project and maintain that version for this particular case.

Thanks!

AgustinCB:
OK, let me give you more detail of the problem:

I'm using this library to do some processing of the image. After that processing, I have a `Mat`. That `Mat` is a face. I want to detect the features of the face as a vector of 68 points.

C++ has a brilliant library called dlib that does that. But it has it's own data structure to represent images, whose format isn't necessarily the same as opencv's. It provides, though, a handy function to convert a `Mat` to its data structure.

I built a basic binding to dlib, but now I'm facing the problem that I don't have how to interconnect this library with my bindings to get the features.

I have some options to solve this problem:

- Rewrite all in C++. I'll have to learn C++ :). And frankly, I'm quite comfortable in rust.
- I make the your `CMat` trait and the `inner` field public, create a bind to that handy function and I pass `inner` in my code.
- I copy a bunch of code from your library into my bindings to dlib and/or I reimplement a lot of what you already have here.
- Something else?

Option 2 works (I tested it from this fork: https://github.com/AgustinCB/cv-rs). However, is horrible and I understand why you don't want to apply that change here.

There's another option: dlib is a huge library and I could probably create more bindings and modify all I'm doing with `cv-rs` to use dlib purely. That would be a lot of work and I wanted to avoid it (hence this thread) but I think at this point it's the solution.

Do you have any other suggestion?

P.S.: Congrats for the library, by the way. Not only is extremely useful and easy to use, it was trivial to me to understand and modify the code being a n00b in Rust.

Pzixel:
Okay, I see your considerations, and you have a huge need in this.

It requires a lot of thinking, but now I see a somewhat way to achieve your goal. We can make `CMat` public, and implement `unsafe impl Into<CMat> for Mat` . In this case you will be able to create CMat from Mat without breaking types as well as not exposing `inner` in public as well as marking it `unsafe` to make sure that no one will do if it's unnecessary.

[...]

PR #77, opened 2018-05-24
Create Into interface for Mat

AgustinCB:
As discussed in #74.

Some caveats: I wasn't able to mark the implementation or the function as `unsafe` without rustc compiling.


src/mat.rs
```diff
@@ -10,11 +10,12 @@ use std::path::Path;
...
 use std::ops::{BitAnd, BitOr, BitXor, Not};
 
+/// The class `CMat` is used as a pointer to represent the Mat opencv structure
 #[derive(Clone, Copy, Debug)]
-pub(crate) enum CMat {}
+pub enum CMat {}
 
 impl CMat {
-    pub fn new() -> *mut CMat {
+    pub(crate) fn new() -> *mut CMat {
```
AgustinCB: I did this to ensure that you don't create `CMat`, but rather get it from a `Mat`. Debatable decision.

Pzixel:
As I said, in Rust taking a pointer to any memory location is safe, it's unsafe when you're trying to deal with it. It's impossible to work with `CMat` in safe context, so people probably won't do it. Great job btw.

src/mat.rs
Pzixel: I completely agree on that

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cv"
-version = "0.2.2"
+version = "0.2.1"
```
Pzixel: Why reverting? I have a strong feeling that every code change have to change a version.

tags
```diff
@@ -0,0 +1,892 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
Pzixel: Please, do not commit autogenerated files.

PR #79, opened 2018-06-05
Dependency fix

Pzixel:
I tried to dogfood with cv-rs and solverd several problems. However, it currently doesn't work as it fails on linking phase with `undefined reference to `cv::img_hash::PHash::create()'`. 

We may switch to bindgen, but we still need higher level abstraction, so it worth to continue adding them.

Looking forward if you could help with the issue.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -47,8 +47,7 @@ fn get_file<'a, T: Iterator<Item = &'a std::io::Result<std::fs::DirEntry>>>(
         ))
     })?;
     let lib = lib.file_name();
-    let lib = lib
-        .into_string()
+    let lib = lib.into_string()
         .map_err(|e| BuildError::new(format!("Cannot convert path '{:?}' to string", e)))?;
     // we expect filename to be something like 'open_world340.lib' or
     // 'open_world.340.dll.a', so we just consider everything after the
--- a/build.rs
+++ b/build.rs
@@ -1,86 +1,120 @@
 extern crate cc;
 
 #[cfg(windows)]
-fn opencv_include() -> String {
-    if let Ok(dir) = std::env::var("OPENCV_DIR") {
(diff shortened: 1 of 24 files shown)


nebgnahz:
On Linux (Travis), we will need to manual add `println!("cargo:rustc-link-lib=opencv_img_hash");` around [line 65 of `build.rs`](https://github.com/Pzixel/cv-rs/blob/1216c32bc6062d89c9545a30282b6d6e9d5b72a8/build.rs#L65).

On Windows (Circle), I am not sure what causes the issue given things are compiled into `opencv_world`. But I suspect a similar issue with linking.

native/phash.cc
```diff
@@ -0,0 +1,23 @@
+#ifndef PHASH_H_
+#define PHASH_H_
```
nebgnahz: The guards should not appear in `.cc` files.

Pzixel: We had it in `utils.cc`. However, you're right.

Pzixel:
Well, currently I have two problems:

1. rustfmt shows the difference although I just updated the chain and used `cargo fmt`.
2. it does compile and runs on linux, but it doesn't on windows. No clue why it's happening.

Pzixel:
One single screenshot to answer 😄 them all

image

build.rs
```diff
@@ -1,77 +1,124 @@
...
+            println!("cargo:rustc-link-lib=opencv_cudaobjdetect");
+        }
     }
 }
 
+#[cfg(windows)]
+use windows::*;
+
+#[cfg(unix)]
+use unix::*;
```
nebgnahz: nice!

nebgnahz:
Cool! LGTM.

PR #80, opened 2018-06-08
Feature/more hash

Pzixel:
Implement all opencv hashes instead of PHash only.


src/hash.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    pub trait HashImpl {
+        fn get_value(&self) -> *mut CHash;
+    }
+}
+
+#[allow(missing_docs)]
+pub trait HashImplInterface: HashImpl {}
+
+/// Basic trait for all OCR types
```
nebgnahz: What is OCR short for? (doesn't seem to be "Optical Character Recognition" to me).

nebgnahz:
> I'd also suggest to remove (incomplete but tested) from the repo header.

Sounds good. Removed.

src/hash.rs
Pzixel: Copy-paste from the same workaround from the OCR module.

PR #81, opened 2018-06-21
Windows intallation fi

Pzixel:
It's about docmuentation right now, but I may add more changes later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,7 @@
 [![Build status][appveyor-image]][appveyor-url]
 [![standard-readme compliant][standard-readme-image]][standard-readme-url]
 
-> This library primarily provides bindings (somewhat idiomatic) and APIs for
-> OpenCV 3.x.
+This library primarily provides idiomatic bindings and APIs for OpenCV 3.x.
 
 [Documentation](https://nebgnahz.github.io/cv-rs/cv/)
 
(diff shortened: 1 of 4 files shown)


Pzixel:
I'm done. Migration to 3.4.1 has broken some steps, rewritten documentation to reflect that. Some batteries such as fixes in scripts are included.

Pzixel:
I see there is duplicate compilations on PR and push. That probably should be fixed in CI settings.

P.S. How are you, globally? Do you have some plans in further project development? We were talking about Go guys, you are busy there?

README.md
```diff
@@ -99,9 +98,9 @@ All possible features are listed below:
 
 ##### Installation steps
 - Create directory `C:\opencv`.
-- Copy files from `.windows` folder there.
+- Copy `.git` and `.windows` folders there (you can run them from the `cv-rs` directory itself, but you may encounter an error that paths are too long)
```
nebgnahz: Why is `.git` folder needed for installation?

Pzixel: Because of submodule restoration.

I think opencv installation probably should be moved into separated crate. And `cv-rs` should just have a link to the repo.

nebgnahz:
> P.S. How are you, globally? Do you have some plans in further project development? We were talking about Go guys, you are busy there?

Much better (time-wise) than before!

About further project plan, at this point, I am not sure where time is best invested: 1. more module wrapper; 2. adopting existing manual wrappers, like gocv's; 3. binding generation. These directions are in an increasing order of complexity and reward.

For 3., modifying OpenCV's python script is one possibility; but last time I checked, it is not that easy to adapt. And the generated bindings are too Python specific. An alternative may be Clang libtooling (haven't explored).

For 1. and 2., it is easy to do it step by step. For 3., as I said earlier, I do not have much time to lead and can only help.

README.md
nebgnahz: Yes, `opencv installation` is better left separate. But what do you mean by "separate crate"?

Pzixel: Sorry, I meant separate repository ;)

Pzixel:
About 3rd bullet: I created [an issue](https://github.com/opencv/opencv/issues/11666), but doesn't look like it get enough attention, unfortunately.

PR #84, opened 2018-07-05
Memory leak fix

### LINKED ISSUE #83: Mat::image_encode is leaking
Hi! Thank you very much for the bindings!

I'm trying to save frames from a video stream as JPEG files, however, as soon as I uncomment the line
`let bytes = mat.image_encode(".jpg", vec![])?;` the app starts consuming more and more memory until it gets OOM-killed.

```rust
extern crate cv;
use cv::*;
use cv::highgui::{Show, WindowFlag};
use cv::imgcodecs::ImageWriteMode;
use std::fs::File;
use std::io::prelude::*;
use std::error::Error;

const WINDOW_NAME: &'static str = "smile";

fn main() -> Result<(), Box<Error>> {
    let capture = videoio::VideoCapture::new(0);
    let window = cv::highgui::highgui_named_window(WINDOW_NAME, WindowFlag::Opengl);

    while capture.is_open() {
        let mat = capture.read();
        if let Some(mat) = mat {
            mat.show(WINDOW_NAME, 50);
//            let bytes = mat.image_encode(".jpg", vec![])?;
//            File::create("img.jpg")?.write_all(bytes.as_slice())?;
        } else {
            break
        }
    }
    Ok(())
}
```

Could you please explain to me what I'm doing wrong or confirm if it is a bug?
Using OpenCV 3.4.1 with GStreamer on rust 1.28.0-nightly.

Pzixel:
Fixes #83


native/imcodecs.cc
```diff
@@ -15,21 +15,17 @@ void* cv_imdecode(const uint8_t* const buffer, size_t len, int flag) {
...
-    uint8_t* buffer = new uint8_t[size];
-    std::copy(buf.begin(), buf.begin() + size, buffer);
-
-    ImencodeResult result;
-    result.status = r;
-    result.size = size;
-    result.buf = buffer;
-    return result;
+    if (r) {
+        CVec<uint8_t> cvec;
```
Pzixel: Not sure if it could lead to use after free. Maybe it should be `CVec<uint8_t>* cvec = nullptr;`

nebgnahz: I don't think it would be use after free. The allocation of the actual buffer is taken care of in `cv_to_ffi`. `CVec` and `COption` are constructed by value. This should work fine.

nebgnahz:
Looks like rustfmt has some upgrades so the CI fails at format checking.

.ci/travis_build_opencv.sh
```diff
@@ -1,20 +1,15 @@
 #!/bin/bash
 set -eux -o pipefail
 
-OPENCV_VERSION=${OPENCV_VERSION:-3.4.0}
-URL=https://github.com/opencv/opencv/archive/$OPENCV_VERSION.zip
-URL_CONTRIB=https://github.com/opencv/opencv_contrib/archive/$OPENCV_VERSION.zip
-OPENCV_BUILD=$(pwd)/opencv-$OPENCV_VERSION/build
-OPENCV_CONTRIB=$(pwd)/opencv_contrib-$OPENCV_VERSION/modules
+OPENCV_VERSION=${OPENCV_VERSION:-3.4.1}
+OPENCV_BUILD=$(pwd)/opencv/build
```
Pzixel: I was keeping version to make sure that caching is working with proper version. That's probably is not required, but worth to be mentioned.

Pzixel:
> Fixing the CI now. There is nothing in /home/travis/build/nebgnahz/cv-rs/opencv-3.4.1 after we use git submodules instead of downloading the zip.

The main idea that `pwd` is `/home/travis/build/nebgnahz/cv-rs/opencv-3.4.1/build`, and it should be there. But instead it tries to build in its parent. It's obviously shouldn't happen, but I don't see why it doesn't start building in *current* directory.

Pzixel:
> Looks like rustfmt has some upgrades so the CI fails at format checking.

That's fine, i will fix those.

Also, do you mind to fix double pr/push checks?

nebgnahz:
Format fix as PR 1 and then rebase for PR 2 is probably better, just in case the memory fix introduces additional bug. But the because format is almost trivial, I am fine with you just combining both into 1 PR.

Pzixel:
I do believe that invalid format occures in newly added code because I actually didn't have llvm on machine where I created the fix. So it's clearly should be one PR.

Issue #78, opened 2018-05-24
Publish into crates.io


Pzixel:
We need to publish this crate into `crates.io`

It the must do thing that wasn't yet accomplished.

What do you think? It still be a toy project that shouldn't appear in places like `crates`? 

Do you have some plans to enhance the library? We did a great job, the library has a single code standard, good readme, nice code conduct and so on... But is it alive?

I understand that you have your own life, but if you don't commit at least once a month then it's probably better to not give people a hope that there is something like that. I neither have an answer what's the right way, It's just a question that appeared suddenly in my head.

nebgnahz:
Hey, @Pzixel, apologies first for the slow response. The busiest time for me has passed, and I can return back to work :) And thanks for handling the PR/issues.

I have thought about publishing on `crates.io`. But hesitated to do so given lack of coverage in terms of functionality.

Also the current process of manually implementing wrappers doesn't feel right to me (it is a good starting point for toy projects though).

I have also re-evaluated the options of automatically generating bindings and failed again. Do you have any luck?

While I was doing some digging into bindgen, I noticed [gocv](https://github.com/hybridgroup/gocv/), which has already implemented all the C wrappers needed for Go to use. So I experimented a bit with bindgen Rust APIs from there. You can find the work in progress here: [code](https://github.com/nebgnahz/rustcv) and [doc](https://nebgnahz.github.io/rustcv/rustcv/).

The generated bindings documented as `opencv_sys` crate is available (https://nebgnahz.github.io/rustcv/opencv_sys/index.html) and accessible as `ffi` module. As a result, adding new functionality is a pure Rust task (so folks who are not familiar with C/C++ don't need to worry about it):

e.g. [median_blur](https://github.com/nebgnahz/rustcv/blob/master/src/imgproc.rs#L281):

```rust
/// Blurs an image using the median filter.
pub fn median_blur(src: &Mat, dst: &mut Mat, size: i32) {
    unsafe { ffi::MedianBlur(src.inner, dst.inner, size) }
}
```

The hope is that there will be one place that all developers can contribute to for the C wrapper (i.e. `gocv`) and both Rust/Go will generate bindings from there.

Since this approach is not compatible with `cv-rs`, I made it a separate repo for now. If it is a viable approach, we can either merge the implementation here or provide a redirect message from this repo there.

What do you think?

Pzixel:
[...]

PR #86, opened 2018-07-09
Redirects opencv internal errors using cvRedirectError

joelgallant:
This is probably not good/idiomatic rust, and I'd appreciate help making it follow the project's style.

Is there a way to unit test this?

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,5 +1,7 @@
 //! Errors for OpenCV bindings
+use std::ffi::CStr;
 use std::path::PathBuf;
+use std::os::raw::{c_int, c_char};
 
 #[derive(Debug, Fail)]
 /// Custom errors that may happen during calls
@@ -16,4 +18,42 @@ pub enum CvError {
     #[fail(display = "Non ascii characters found in string: {:?}", _0)]
     /// Indicates that string contains non ascii characters
     UnicodeChars(String),
+}
+
+extern "C" {
+    fn cvRedirectError(callback: unsafe extern fn (status: c_int,
+                                                   func: *const c_char,
+                                                   msg: *const c_char,
+                                                   file_name: *const c_char,
+                                                   line: c_int) -> c_int);
+}
+
+// handle that can be set by user
+static mut GLOBAL_CATCH: Option<fn (status: i32) -> i32> = None;
+
+unsafe extern "C" fn err_handler(status: c_int,
+                          func_name: *const c_char,
+                          err: *const c_char,
+                          file_name: *const c_char,
+                          line: c_int) -> c_int {
+    let to_str = |p: *const c_char| -> &str { CStr::from_ptr(p).to_str().unwrap_or("") };
+
+    println!("opencv err: {} - {} in {}:{} (status: {})",
+             to_str(err),
+             to_str(func_name),
+             to_str(file_name),
+             line,
+             status);
+
+    if let Some(catch) = GLOBAL_CATCH {
+        catch(status)
+    } else {
+        panic!("failed in opencv function");
+    }
+}
+
+/// tells opencv to print out errors
+pub unsafe fn catch_opencv_errors(cb: Option<fn (status: i32) -> i32>) {
+    cvRedirectError(err_handler);
+    GLOBAL_CATCH = cb;
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,6 +34,7 @@ pub mod text;
 pub mod video;
 pub mod videoio;
 
+pub use errors::catch_opencv_errors;
 pub use core::*;
 pub use mat::*;
 


joelgallant:
I agree re: one global handler - it feels dirty but I wasn't sure what the appropriate way to handle it is.

I required this because a fair amount of opencv functions will terminate on assertions, and not output debug messages or anything when the program crashes.

It seems that it's doing [this](https://github.com/opencv/opencv/blob/2da96be217ab437de854aca7f7670c2048fb554b/modules/core/src/system.cpp#L1002), which panics underneath rust which doesn't output any useful messages.

How it's done inside of opencv: https://github.com/opencv/opencv/blob/2da96be217ab437de854aca7f7670c2048fb554b/modules/core/src/system.cpp#L897

joelgallant:
It would be extremely onerous for a single handler to dispatch errors anywhere inside of this library, somehow (I have no idea how that might even work).

Pzixel:
> I required this because a fair amount of opencv functions will terminate on assertions, and not output debug messages or anything when the program crashes.

As far as I see it's disabled by default and never gets set to true:

```cpp
static bool breakOnError = false;
```

So you shouldn't be affraid of unexpected panic  - switch is never enabled.

I actually implemented a wrapper that handles an exception if it occurs and transforms into rust Result type, e.g. see https://github.com/nebgnahz/cv-rs/blob/master/native/common.h#L68 and https://github.com/nebgnahz/cv-rs/blob/master/native/text.cc#L30

It's true that in most places we are just panicing, but it happens because cpp code doesn't handle exceptions properly and they raise panic on Rust side.

So I think the better approach is just find the places where exception may be raised and wrap them into `FromFunction` so they would be correctly propagated in Rust code.

> It would be extremely onerous for a single handler to dispatch errors anywhere inside of this library, somehow (I have no idea how that might even work)

I personally thing the opposite: all kinds of something global should not exist until absolutely necessary. I think that local `Result`'s are much better, because you have the entire context required to handle the error. With global handler you could just log it somewhere, which is not really helpful. And if we're talking about some production-ready code, you have to have some kind of configurable DI and so on so it's become pain to work with such kind of global objects.

joelgallant:
> all kinds of something global should not exist until absolutely necessary

Totally agree, sorry if it sounds like I want this, I'd much rather deal with result types. As far as I could tell, opencv exceptions/errors that are thrown in CV_Assert aren't catcheable when called via the C API (like they could be with c++ exceptions). If I'm wrong about that, that's awesome.

I guess the idea is to wrap every native function like that in a c++ closure, and utilize the result type?

Pzixel:
> As far as I could tell, opencv exceptions/errors that are thrown in CV_Assert aren't catcheable when called via the C API (like they could be with c++ exceptions).

That's probably true, but the main idea here that we don't use C API to call OpenCV. It's our own wrappers (see `native` folder) that could do everything, including transforming exceptions as I have shown. But I generaly don't believe in uncatchable panics instead of exceptions - that could be a disaster for the whole ecosystem! :) I believe OpenCV devs are working with such a big project and make such and just couldn't do that.

> I guess the idea is to wrap every native function like that in a c++ closure, and utilize the result type?

Yes, it's the main idea, except for places where exception could never occur. That's what I have written above:

> So I think the better approach is just find the places where exception may be raised and wrap them into `FromFunction` so they would be correctly propagated in Rust code.

joelgallant:
My apologies for not taking more time to understand this repo before confusing you with this PR, I now understand this much better.

With OpenCV, exceptions could crop up anywhere. Is there a guarantee that _all_ ffi functions are wrapped in a catch-all? Otherwise, is it of any use to use some variation of this PR? (without the global catcher maybe) I just spent some time with this library and ended up running into some, what look like on the surface, segfaults, which were actually assertions internal to OpenCV that were not caught. I'm sure the idea is to catch all of these and create appropriate results, but in lieu of that, maybe utilizing this redirection would help users of the library?

Just my 2 cents - if you prefer the fix-them-as-they-come approach that's cool too, this just helps in my debugging (as I need to implement more functions than are currently supported, so I am running into this more).

Pzixel:
> My apologies for not taking more time to understand this repo before confusing you with this PR, I now understand this much better.

That's perfectly fine, I'm glad you found time to dive deeper into it.

> With OpenCV, exceptions could crop up anywhere. Is there a guarantee that all ffi functions are wrapped in a catch-all?

If you wrap them all - yes, you can 😄 However, it's not done yet. I just wrapped functions I personally worked with, and where I personally was getting panic instead of errors. I just didn't get enough time to handle all places. You can easily wrap all calls, but it's not perfect since some of them just couldn't rise any exception. In this case it's just complicates API and usage experience without any benefit. Another point is that we were thinking about code generation instead of current manual wrapping approach. Because of that, I personally maintain the project fixing bugs people find, but I don't add much new value. There is multiple discussion about it, e.g. https://github.com/nebgnahz/cv-rs/pull/67#issuecomment-370133765 and #78. So I'm not sure about how we should proceed.

> Just my 2 cents - if you prefer the fix-them-as-they-come approach that's cool too, this just helps in my debugging (as I need to implement more functions than are currently supported, so I am running into this more).

I'd appreciate any work you want to share.

PR #87, opened 2018-07-10
Adds canny edge detection

joelgallant:
Attached to Mat instead of making it a function with an `src, dst` signature, seemed like that is the preference in this lib.

### DIFF
--- a/native/imgproc.cc
+++ b/native/imgproc.cc
@@ -93,10 +93,9 @@ void cv_compare_hist(cv::Mat* first_image, cv::Mat* second_image, int method, Re
         [first_image, second_image, method]() { return cv::compareHist(*first_image, *second_image, method); });
 }
 
-void cv_canny(cv::Mat* image, cv::Mat* edges, double threshold1, double threshold2, int aperture_size, bool l2_gradient, Result<cv::Mat*>* result) {
-    *result = Result<cv::Mat*>::FromFunction([image, edges, threshold1, threshold2, aperture_size, l2_gradient]() { 
+EmptyResult cv_canny(cv::Mat* image, cv::Mat* edges, double threshold1, double threshold2, int aperture_size, bool l2_gradient) {
+    return EmptyResult::FromFunction([image, edges, threshold1, threshold2, aperture_size, l2_gradient]() { 
         cv::Canny(*image, *edges, threshold1, threshold2, aperture_size, l2_gradient);
-        return edges;
     });
 }
 }
--- a/native/imgproc.cc
+++ b/native/imgproc.cc
@@ -92,4 +92,11 @@ void cv_compare_hist(cv::Mat* first_image, cv::Mat* second_image, int method, Re
     *result = Result<double>::FromFunction(
         [first_image, second_image, method]() { return cv::compareHist(*first_image, *second_image, method); });
 }
+
+void cv_canny(cv::Mat* image, cv::Mat* edges, double threshold1, double threshold2, int aperture_size, bool l2_gradient, Result<cv::Mat*>* result) {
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -54,10 +54,10 @@ struct CResult<T: Copy> {
     error: CDisposableString,
 }
 
-impl<T: Copy> Into<Result<T, String>> for CResult<T> {
+impl<T: Copy, V: Into<T> + Copy> Into<Result<T, String>> for CResult<V> {
     fn into(self) -> Result<T, String> {
         if self.error.value.is_null() {
-            Ok(self.value)
+            Ok(self.value.into())
```
joelgallant: Unnecessary change, but I think this makes sense?

Pzixel: Thank you for the PR on the first place.

I follow yagni principle so I think we don't need to add it until it's not required somewhere.

Pzixel:
The format checking is failing. You are probably uses invalid `cargo fmt`. You need have exactly same version, because they are a bit incompatible with each other. I will create a PR with fmt update because it's a shame.

Pzixel:
Merge latest master. It should do the trick

Pzixel:
CI currently uses `rustfmt 0.8.2-nightly (5e59925 2018-07-02)`. You can check your version via `cargo fmt --version`. If your version differs then your code may not pass.

joelgallant:
Oops, should have checked first. Re-pushed with correct fix. (and is rebased on `master`)

Pzixel:
Okay, now `cargo fmt` check passed, and it failed on clang one (we are using clang-format-5 if I remember correctly). And if you don't mind I'l post some some review comments.


P.S. Sorry for this much iterations, I think I just had to link out [contributing readme](https://github.com/nebgnahz/cv-rs/blob/master/CONTRIBUTING.md) which contains all required links.

native/imgproc.cc
```diff
@@ -92,4 +92,11 @@ void cv_compare_hist(cv::Mat* first_image, cv::Mat* second_image, int method, Re
     *result = Result<double>::FromFunction(
         [first_image, second_image, method]() { return cv::compareHist(*first_image, *second_image, method); });
 }
+
+void cv_canny(cv::Mat* image, cv::Mat* edges, double threshold1, double threshold2, int aperture_size, bool l2_gradient, Result<cv::Mat*>* result) {
+    *result = Result<cv::Mat*>::FromFunction([image, edges, threshold1, threshold2, aperture_size, l2_gradient]() {
```
Pzixel: You return `edges` as output parameter and as function result. I don't think it's a good thing, so I propose to add CEmptyResult which works the same way as `CResult<T>` but that doesn't have any result (I think, it's better than having CResult<Some_ignored_type_parameter>). E.g. something like:

```cpp
struct EmptyResult {
    CDisposableString error;

    static EmptyResult FromFunction(std::function<void()> function) {
        char* error = nullptr;
        try {
            function();
        } catch (cv::Exception& e) {
            const char* err_msg = e.what();
            auto len = std::strlen(err_msg);
            error = new char[len + 1];
            std::strcpy(error, err_msg);
        }
        return Result{CDisposableString{error}};
    }
};
```

So signature of the function could be `EmptyResult cv_canny(...)`

tests/test_imgproc.rs
```diff
@@ -70,3 +70,10 @@ fn get_image_histogram(path: &'static str) -> Mat {
     let image = image.normalize(0.0, 1.0, NormType::MinMax);
     image
 }
+
+#[test]
+fn canny_edge_detection() {
```
Pzixel: Do you mind adding test for error case as well?

joelgallant:
Okay, I've updated this PR with a failing test and returning an EmptyResult

Pzixel:
sorry, my new year finally ended :)

Thank you for your PR.

Issue #107, opened 2019-03-04
Allow libs to be built independently on Windows


vadixidav:
I built the associated OpenCV on Linux and Windows (using my own CMake command, since the included powershell script was complaining about path lengths no matter what I did). On Windows, I didn't get an `opencv_world` lib. I assume this is due to not specifying `INSTALL_CREATE_DISTRIB=ON`.

I think that there should be an env var to support both build modes. I can add this functionality if desired. Is this acceptable? Also, I don't have a good name for the flag. What should it be?

Pzixel:
Hello.

I see no harm in this behaviour, it would be even more consistent since linux actually doesn't build a `world`.

Flag may be called like "combine_world" or something. Have no good ideas now too :)

vadixidav:
@Pzixel I actually came up with a better idea. I can just search if `world` is there and if it isn't I can try linking to the individual libs. Is this fine?

vadixidav:
This should be solved by #117. When using the system installation on Linux it uses pkg-config and on Windows it just links every lib in the libs folder.

PR #108, opened 2019-03-05
fix windows build.rs section to try all-in-one and separate libs

vadixidav:
This solves #107.


Pzixel:
All commits gets squashed so I can merge them separately, but you'l have to fix conflicts. Let's stick with single #116 issue

PR #111, opened 2019-03-05
Fix lints

### LINKED ISSUE #109: Fix clippy lints
I would like to go through and just slash through all the Clippy lints. Some things are breaking changes. Is this fine? For instance, an example of a breaking change would be `DescriptorMatcher::add` which takes a `&Vec<&Mat>` when it should take a `&[&Mat]` or `IntoIterator<Item=&Mat>`.

vadixidav:
Closes #109.


Pzixel:
Duplicate of #116 ?

PR #117, opened 2019-03-07
Refactor

vadixidav:
This branch attempts to refactor the library into a sys crate. Currently the sys crate is building, but the existing `native` folder is not as useful anymore now that there are actually native C++ types exposed to Rust. I might start by trying to get it all working with the previous bindings first before progressing to remove all the extra fake types and pointers.


vadixidav:
CUDA is also completely untested for the time being.

Pzixel:
I propose to not remove constructors since having builders only is not a good way.

vadixidav:
@Pzixel There is a clippy lint for having too many args. I agree that the original had too many args. I can choose a few args that most people will want to set or I can just make a default constructor. I feel like just having a default may be sufficient here. Does that sound fine?

vadixidav:
Also, I am going to take off work today to finish this sys crate refactoring and add additional functionality to the crate (in a follow-up PR).

Pzixel:
> There is a clippy lint for having too many args. I agree that the original had too many args. I can choose a few args that most people will want to set or I can just make a default constructor. I feel like just having a default may be sufficient here. Does that sound fine?

This makes usecase when you know all arguments but have to use build because there is no other API less pleasant.

You either leave all 100500 arguments as they are or create a struct that encapsulate all of them, but you shouldn't just remove it with "use builders, guys".

> I can choose a few args that most people will want to set

And this one fits into builder fine. You can create `SIFTBuilder::create_jpg_processor()` with best params that handles JPG, for example.

vadixidav:
Some of the things I am doing I will improve. For instance, currently there is some unsafe code that can be avoided (or isolated to a single instance). Before marking the PR ready for review I will resolve those issues.

vadixidav:
I finally got `cv-sys` to build with C++ bindings now. There is no need to use extern "C" in the headers or C++ files. We can also expose methods to Rust provided certain conditions. Many of the opencv things we are interacting with can even have bindings generated for them, so we might be able to gradually erode away the wrapper entirely.

The last bit of work required is now just updating the `cv` crate to use the new bindings which have changed a little. I may need to expose a few more opaque types from OpenCV as well. Before marking ready for review I will turn all the enums I made into opaque types passed to bindgen so that they are generated automatically. I will also ensure that everything works with a prebuilt OpenCV as well using a feature on the sys crate.

I haven't tested the CUDA stuff at all, and I know it wont work. I hope that once I have everything sorted out that either I can get CUDA to work or somebody else can help with that.

vadixidav:
I have found that, at least on Debug Windows static builds (I haven't gotten to others yet), `canny_edge_detection` seems to have a segfault. I don't know if I want to track down why that happens quite yet, but it probably needs to be tracked in this repository and also upstream in OpenCV if it turns out to be a bug in OpenCV.

vadixidav:
Okay, now that all the functionality is working again (only on Debug Windows), I need to make sure it works on all builds (Windows, Linux, etc). I also need to make sure that CUDA works. I don't have a Mac, but I am not sure we ever supported that. Finally, I need to make it work for a prebuilt OpenCV with a feature flag.

- [x] Windows Debug
- [ ] Windows Release
- [ ] Linux Debug
- [ ] Linux Release
- [ ] Windows CUDA
- [ ] Linux CUDA

vadixidav:
There is a place in the bindings where `bindgen` uses the type `u64` for `MouseCallback`. This is a function pointer, so there is a problem with the type changing from one platform to the next. We need to either make sure the binding generates a `usize` or a `fn(...)` type so that we dont have issues on non-64-bit platforms. Alternatively, we can use the pointer size in some `cfg` statements.

vadixidav:
Linux Debug and Release is now good to go with the refactor. I am now going to perform some clean up and also generate bindings to a prebuilt OpenCV so that works. I still need to make CUDA work. Give me some time to ensure that is still working properly.

Also, I tried building on a machine with Visual Studio 2015 and encountered some issues. Ideally we would like to fix that (particularly for CUDA support). I will investigate if that is in the `bindgen` phase or the `cmake` phase. If it is the `cmake` part of the build script, then it's not a blocker since they can just use a prebuilt OpenCV. If it is a problem with the `bindgen` part (most likely) then since the new bindings system relies on bindgen there will be issues.

vadixidav:
Confirmed today that there are no issues on Visual Studio 2015. The issue encountered was that Clang was not installed. After Clang was installed, bindgen worked as expected.

vadixidav:
Seems that OpenCV fails to build with Clang (it works with g++). OpenCV master seems to have this issue fixed, but I wont make this a blocker.

Pzixel:
Cool.

I only suggest to use bindgen CLI instead of package. I heard tons of stories about problems because of it. Storing CLI-generated source in repo is much better than having build.rs with bindgen description. 

I can be in more detail if you wish, but the general suggestion is this.

vadixidav:
@Pzixel I was considering doing that, but hadn't yet made a decision. It might be best to do like you say and add the bindings to the repository. That would eliminate issues with bindgen giving different results in different scenarios (like different versions of libclang or different C++ headers). I will look into that after I make everything work with the system OpenCV install (behind a feature flag). It might actually be good to have a feature to generate the bindings optionally just so that we have the bindgen configuration tracked in revision control and tested in CI.

- [ ] Add generated bindings into repository and make generating them optional.

vadixidav:
The refactoring is complete and all issues brought up in this and the other PR I closed have been dealt with. Moving on from here, we can generate native bindings directly to C++ wrapper code, making the creation of bindings simpler. We can also potentially start whitelisting more of the bindings from the `cv::` namespace and get rid of the wrappers entirely, but that is outside of the scope of this PR.

I would like you @Pzixel to try and enable the `cuda` feature (via `--features cuda,system`) at the command line with a CUDA-enabled OpenCV install. It almost certainly wont work, but I haven't been able to get OpenCV to build with CUDA, even on my CUDA enabled machine, because it refuses to install the Visual Studio integration. I need somebody to help fix up the `cuda` bindings.

As for the `text` and `tesseract` features, those may be untested as well.

I am going to boot into Debian and make any remaining fixes I need to so that it works on there. That should be the last thing that needs to be done.

Pzixel:
@vadixidav did you use `msvc_1_install_CUDA.ps1`? It should configure everything for you. The only requirement is VS2015 since VS2017 doesn't work with CUDA.

vadixidav:
@Pzixel Unfortunately it didn't work on my VS 2015 machine. The manual install also failed. Other people in my office have had similar issues where it fails VS integration. I don't see myself easily being able to access a cuda machine on 2015 in the near future. It also fails if you have multiple Visual Studio installations, so I couldn't do it on my laptop with VS 2017 without uninstalling VS 2017. Plus, that laptop uses an eGPU to access a GTX 1080, so I don't know how CUDA would interact with me plugging the GPU in while its running on Windows or Linux.

vadixidav:
Working on making this work on Debian right now.

vadixidav:
Unfortunately I had to add generated bindings for Linux and Windows. It seems that names are mangled differently between gcc and MSVC, so we cant rely on common name-mangling. I wish we could generate bindings at build since it barely takes any time at all, but I wouldn't want downstream users to have to install Clang. Hopefully eventually rustup will package `clang` and `libclang` like it does for other llvm utilities.

vadixidav:
Using a pre-installed OpenCV on Linux has issues. I am going to try and wrangle the dependencies with pkg-config in all instances.

vadixidav:
I will see if I can get CUDA working on my Debian install on my laptop. If I can, I might be able to get everything through CI. I am going to have to generate another set of bindings for CUDA-enabled OpenCV as well.

vadixidav:
At a Linux CUDA machine now to test.

vadixidav:
@Pzixel I think we should consider depending on libclang to be installed and have bindings always generated. Unlike C, [C++ makes no guarantee of name mangling stability](https://stackoverflow.com/questions/10127982/why-is-name-mangling-not-standardized) even between compiler versions. We could have to support an inane amount of generated bindings. We could have had one set of bindings before with the C++ -> C -> Rust binding process, but now that we are going C++ -> Rust, it seems that generating bindings to C++ might have to be a requirement. What are your thoughts?

Pzixel:
@vadixidav You could try removing `vs2017` check. I introduced it becuse in a time CUDA didn't work well with VS2017. I think they might already fix it so compiling with 2017 may work.

>  Plus, that laptop uses an eGPU to access a GTX 1080, so I don't know how CUDA would interact with me plugging the GPU in while its running on Windows or Linux.

Worth trying anyway, imo

About your mangling question I think if nomangle may work. I don't see any solution working in all cases. We either have unstable bindings or have to pregenerate them and deal with mangling issues or something. 

---

P.S. I'm planning sending https://github.com/opencv/opencv/issues/11666 to the https://github.com/opencv/opencv/wiki/GSoC_2019 and see if it will work. It won't benefit immediately, but we could switch to stable C ABI later if needed.

vadixidav:
@Pzixel Oh, that would be awesome if somebody worked on that. Perhaps make a tracking issue in the repository once its up.

Pzixel:
Btw I don't think replacing opaque enums with `c_void` is a good thing.

vadixidav:
@Pzixel I agree. The reason it is like this currently is because if the types are not blacklisted then bindgen generates methods for them (from C++), but if they are blacklisted then it replaces pointers to them with `*mut c_void`. I would like to create wrapper types on the C++ side or attempt to generate bindings directly from the C++, but that would require yet further refactoring that I didn't want to do yet. Bindgen might let me use `typedef Blah cv::Blah` in the `cvsys` namespace, then I can use `*mut cvsys_Blah` in the code. Let me try doing that. Right now I am trying to fix the `text` feature.

vadixidav:
Alright, now everything is typed. I created types on the C++ side to get around limitations with templating in bindgen. Now bindgen generates types on the Rust side that correspond to the C++ side and all pointers are now typed.

I would now like to revisit Windows one last time to make sure everything still builds on Windows in all configurations. I will see if I can get `text` and `tesseract` to work on Windows.

I would really appreciate it if someone else could try testing CUDA support on Windows. I haven't had any success with it so far personally, but I am sure the build script has some issues to work out in that scenario.

I will leave a comment on here after everything seems to be working fine on Windows. If it all works then I think this will be ready to merge.

vadixidav:
Alright, aside from CUDA on Windows, I feel pretty confident about this build. I have tested it on multiple machines in every configuration.

We can always clean Windows CUDA support up if we need to later.

The last step now is to just make sure CI passes, which I will now await.

vadixidav:
I got CUDA to work on my laptop (CUDA 10 with Visual Studio 2017). Apparently it does work with Visual Studio 2017 after all now with CUDA 10. I will do some tests on my laptop with my eGPU and make sure the opencv build works there.

I am still trying to get things working on CI. You will know when it turns green that things are good to go.

vadixidav:
There is some weird issue with IlmImf on the Ubuntu 16.04 build server. I don't have a 16.04 box, so this might take me some time to investigate.

vadixidav:
After this last build, I think it is time for me to take a break from this. I want to actually start adding APIs and not refactoring, even though there is probably more work to be done. I also have other open-source work to do rather than wrangle OpenCV builds.

So, I think it should be ready to merge after this. I just updated the README.md and the travis to reflect it. It builds on several versions of Ubuntu and Debian. It works on Windows. It works with and without CUDA. It works with tesseract. I did a lot of things to get it to work everywhere, so I hope it is good enough to merge now. Despite my best efforts, trying to wrangle a build of OpenCV was not easy, and I hope people will try to address the kinks in the future (like lack of OS X support). I really still don't feel super confident about the build, but it has worked on many machines, so hopefully that is sufficient for now.

Now after this I can move onto adding all the features I need to the package.

vadixidav:
@Pzixel Feel free to review my changes. I am ready for this to be merged.

Pzixel:
Okay.

It will also take a while so feel free to use `git` dependency with branch specification if you need it.


P.S. I wonder why there is no appveyor builds. Did you removed `yml`?
P.P.S. Please, fix this minor PR conflict.

vadixidav:
@Pzixel I am not sure why AppVeyor didn't build that last commit. The last two succeeded, but then it just ignored the last one. After I resolve the conflict, it should work.

vadixidav:
@Pzixel Both CI ran this time and are passing.

.ci/install_cuda.sh
```diff
@@ -1,18 +0,0 @@
-#!/bin/sh
```
Pzixel, line 1: So everything was moved to the `cmakelists`, right?

.travis.yml
```diff
@@ -1,8 +1,7 @@
 env:
   global:
-    - MAKEFLAGS="-j 4"
```
Pzixel, line 3: Don't you think it worth having multithreaded build?

Cargo.toml
```diff
@@ -11,17 +10,18 @@ documentation = "rust-vision.s3-website-us-west-2.amazonaws.com"
...
+itertools = "0.8.0"
 
 [features]
-cuda = []
-text = []
-tesseract = ["text"]
+cuda = ["native/cuda"]
+text = ["native/text"]
+tesseract = ["text", "native/tesseract"]
+system = ["native/system"]
```
Pzixel, line 26: What this feature stands for?

README.md
```diff
@@ -21,88 +21,81 @@ This library primarily provides idiomatic bindings and APIs for OpenCV 3.x.
...
-- Installed CMake x64 ([download link](https://cmake.org/download/)).
-- Installed MinGW ([download link](https://sourceforge.net/projects/mingw-w64/files/latest/download)). Choose architecture `x86_64` during installation.
-
-##### Installation steps
-- Create directory `C:\opencv`.
-- Copy `.git` and `.windows` folders there (you can run them from the `cv-rs` directory itself, but you may encounter an error that paths are too long)
-- Run powershell console as administrator in `c:\opencv`.
-- Run `PowerShell -NoExit -File .\.windows\mingw_build_OCV.ps1 -MinGWPath "C:\Program Files\mingw-w64\x86_64-7.2.0-posix-seh-rt_v5-rev1\mingw64\bin"` (your path may be different).
-- Wait until installation finishes. Now you have properly configured OpenCV.
+You must have `cmake` installed and it otherwise just works. It will build
```
Pzixel, line 95: I think it worth leave CMake/VS links as they are, it's more user-friendly.

cv-sys/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cv-sys"
+version = "0.1.0"
+authors = ["Geordon Worley <vadixidav@gmail.com>", "Ben Zhang <benzh@cs.berkeley.edu>"]
```
Pzixel, line 4: I'd like to be added as well 😄 . Or maybe you could leave yourself alone since it was done completely by you.

cv-sys/build.rs
```diff
@@ -0,0 +1,366 @@
+extern crate bindgen;
+extern crate cmake;
```
Pzixel, line 2: It's better to use RUST-2018 import style indead of `extern`.

cv-sys/build.rs
```diff
@@ -0,0 +1,366 @@
...
+    // Finally generate the bindings.
+    let bindings = bindings.generate().expect("bindgen was unable to generate bindings");
+
+    // Write the bindings in the build directory.
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    bindings
+        .write_to_file(out_path.join("bindings.rs"))
+        .expect("Couldn't write bindings!");
+
+    Ok(())
```
Pzixel, line 365: It's better to separate 300-lines main method to something simpler. It could easier to maintain this way.

cv-sys/modules/core/include/opencv2/core/cv_cpu_helper.h
```diff
@@ -0,0 +1,274 @@
+// AUTOGENERATED, DO NOT EDIT
```
Pzixel, line 1: What's the tool for autogeneration? Shouldn't it be added to gitignore?

cv-sys/native/flann.hpp
```diff
@@ -0,0 +1,6 @@
+#ifndef CV_RS_FLANN_H
```
Pzixel, line 1: what's the purpose of these empty headers?

cv-sys/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+#![allow(clippy::all)]
+
+use std::ffi::CStr;
+use std::iter::FromIterator;
+
+// Include bindings.
+include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+// Create opaque types.
+pub enum cv_BOWKMeansTrainer {}
```
Pzixel, line 14: You removed `cv` prefix from everywhere, don't you think to remove it here as well? I think it's better to move to the `mod` if you expect name clashing.

cv-sys/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+pub enum cv_TermCriteria {}
+pub enum cv_VideoCapture {}
+pub enum cv_VideoWriter {}
+pub enum cv_DMatch {}
+pub enum cv_KeyPoint {}
+pub enum cv_Point {}
+pub enum cv_Rect {}
+pub enum std_string {}
+
+impl Clone for cvsys_Point2i {
```
Pzixel, line 28: Don't you think it's better to apply implementation via macro? I think there should be ways to do it with bindgen or at least decorate externally.

cv-sys/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+impl<T> Into<std::result::Result<T, String>> for cvsys_Result<T> {
+    fn into(mut self) -> Result<T, String> {
+        if unsafe { self.error.is_str() } {
+            let ret = Err(unsafe { CStr::from_ptr(self.error.get_str()) }
+                .to_str()
+                .expect("got non-UTF8 error string from OpenCV")
+                .to_owned());
+            unsafe {
+                self.error.drop_cpp();
+                std::mem::forget(self);
```
Pzixel, line 97: I'm not sure about `mem::forget` here. Regular drop should work fine so we don't need to call `drop_cpp` and `forget` externally.

cv-sys/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+impl<T, U> Into<Vec<U>> for cvsys_CVec<T>
+where
+    T: Into<U> + Clone,
+{
+    fn into(self) -> Vec<U> {
+        Vec::from_iter(
+            (0..self.size)
+                .map(|n| unsafe { &*self.array.add(n) })
```
Pzixel, line 123: Why we are calling `add` method (which should mutate the storage according ot its name) to get items from the vec? Shouldn't it be something like `get`?

examples/sift.rs
```diff
@@ -0,0 +1,68 @@
...
+    let (green_keypoints, _) = sift.detect_and_compute(&green, &mask);
+    let (blue_keypoints, _) = sift.detect_and_compute(&blue, &mask);
+
+    for (kps, (c1, c2)) in
+        [red_keypoints, green_keypoints, blue_keypoints]
+            .iter()
+            .zip(&[(&green, &blue), (&blue, &red), (&green, &red)])
+    {
+        for kp in kps {
+            let x = kp.point.x as c_int;
```
Pzixel, line 43: use regular `i32` instead of all `c_int`. See http://answers.opencv.org/question/184248/does-opencv-runs-on-platforms-with-different-int-size/  and #58 . ~I believe I had a gurard that checks that `c_int` is actually 32bit integer.~ No, I only suggested in in the PR. However I don't recall why I didn't implement it. May worth adding it.

src/core.rs
```diff
@@ -30,6 +23,32 @@ pub struct KeyPoint {
     pub class_id: c_int,
 }
 
+impl From<native::cvsys_KeyPoint> for KeyPoint {
+    fn from(n: native::cvsys_KeyPoint) -> Self {
+        Self {
+            point: n.pt.into(),
+            size: n.size,
```
Pzixel, line 30: you probably could use spread operator here

src/core.rs
```diff
@@ -62,6 +81,28 @@ impl Scalar {
     }
 }
 
+impl From<native::cvsys_Scalar> for Scalar {
```
Pzixel, line 84: I've discussed this matter with several Rust full-time developers.

They suggest either just reexport `cvsys` struct or implement newtype over scalar with getters/setters. Unnessessary copying is considered a bad practice.

I think the former is the best approach.

src/features2d/descriptor_matcher.rs
```diff
@@ -64,75 +49,84 @@ impl DescriptorMatcherType {
 
 /// Type for matching keypoint descriptors
 #[derive(Debug)]
-pub struct DescriptorMatcher {
-    value: *mut CDescriptorMatcher,
+pub struct DescriptorMatcher<'a> {
+    value: *mut native::cvsys_DescriptorMatcher,
+    phantom: PhantomData<&'a ()>,
```
Pzixel, line 54: I'm not sure it's a good idea. We are producing unbound lifetime in the `new` method so it's quite useless.

src/features2d/descriptor_matcher.rs
```diff
@@ -64,75 +49,84 @@ impl DescriptorMatcherType {
...
             );
         }
-        matches.unpack()
+        matches.into()
     }
 
     /// Finds the k best matches for each descriptor from a query set.
     pub fn knn_match(&self, query_descriptors: &Mat, k: usize) -> Vec<Vec<DMatch>> {
-        let mut matches = CVec::<CVec<DMatch>>::default();
+        let mut matches: native::cvsys_CVec<native::cvsys_CVec<native::cvsys_DMatch>> = unsafe { std::mem::zeroed() };
```
Pzixel, line 120: I'm not sure it will work hte right way. Previosely I was initing `CVec` with nullptr without actually allocating any memory. Could you elaborate on than?

src/features2d/descriptor_matcher.rs
```diff
@@ -64,75 +49,84 @@ impl DescriptorMatcherType {
...
         unsafe {
-            cv_matcher_knn_match(self.value, query_descriptors.inner, k as c_int, &mut matches);
+            native::cvsys_matcher_knn_match(self.value, query_descriptors.inner, k as c_int, &mut matches);
         }
-        matches.unpack()
+        let match_conv = |matches: native::cvsys_CVec<native::cvsys_CVec<native::cvsys_DMatch>>| {
```
Pzixel, line 124: I think it's better to import some `native` types to make it more readable.

src/features2d/mser.rs
```diff
@@ -3,167 +3,157 @@ use core::*;
...
+/// - min_area: 60,
+/// - max_area: 14400,
+/// - max_variation: 0.25,
+/// - min_diversity: 0.2,
+/// - max_evolution: 200,
+/// - area_threshold: 1.01,
+/// - min_margin: 0.003,
+/// - edge_blur_size: 5,
+#[derive(Debug, Copy, Clone)]
+pub struct MserBuilder {
```
Pzixel, line 58: It's not a builder since builder is working with optional fields. For example there is no way to determine if user uses `5` intentionally or it comes from default. And it could create a regression if/when opencv changes defaults and we do the same.

src/features2d/surf.rs
```diff
@@ -103,11 +82,18 @@ impl Into<SURF> for SURFBuilder {
 
 impl Feature2D for SURF {
     fn detect_and_compute(&self, image: &Mat, mask: &Mat) -> (Vec<KeyPoint>, Mat) {
-        let mut keypoints = CVec::<KeyPoint>::default();
-        let descriptors = CMat::new();
         unsafe {
-            cv_surf_detect_and_compute(self.value, image.inner, mask.inner, &mut keypoints, descriptors, false);
+            let mut keypoints: native::cvsys_CVec<native::cvsys_KeyPoint> = std::mem::zeroed();
```
Pzixel, line 86: +1 Nullptr/alloc question here

src/highgui.rs
```diff
@@ -118,19 +139,41 @@ pub enum MouseEventType {
     MouseHWheel = 11,
 }
 
+impl From<c_int> for MouseEventType {
```
Pzixel, line 142: Do you mind reference `num_derive` and detive `FromPrimitive` trait instead?

Pzixel (changes requested):
Tremendous work! Thank you for all your time. I spend last 4 hours only reviewing it.

I could miss something so I'l review it from the scratch once more later.

I left some sugesstions and questions below.

Pzixel:
@nebgnahz what's your opinion?

vadixidav:
I have been busy with some other things (`hwt` crate and `akaze` crate), but I will address some of the issues brought up here now. Hopefully I can find some time to fix these soon before I diverge too much from master.

src/highgui.rs
vadixidav: One potential problem is that `c_int` could be an i32 or an i64. I couldn't think of another simple way to solve that issue and settled for this for now.

src/features2d/surf.rs
vadixidav: This code is not ideal (I shouldn't be using `std::mem::zeroed`), but on the C++ side the pointer is filled in and when the `CVec` is dropped the C++ side cleans up the memory if the pointer is set. In the future, or possibly now, we should create a safe `new` method. This was just a quick hack.

src/features2d/mser.rs
vadixidav: The defaults are set in the `Default::default()` implementation. Rust has no default parameters or anything of the like. We could pass the options into C++, but since these are arguments (in C++), we must specify all of the arguments that come before one that we specify. This means that we must provide a default value and cannot rely on the ones from OpenCV. For this reason, all the values are filled and they are all passed every time on the C++ side.

If there is some advantage to changing the fields to be an `Option` (such as to optionally do something on the C++ side), please let me know and I can try to reconcile that situation.

src/features2d/descriptor_matcher.rs
vadixidav: Alright, I will see what I can do to improve the readability.

- [ ] Add some `use` statements to import types from `native`

src/features2d/descriptor_matcher.rs
vadixidav: This code allows C++ to perform the allocation, which sets the pointer. On Drop the `CVec` frees the memory. The implementation is in `cv-sys/src/lib.rs`.

cv-sys/Cargo.toml
vadixidav: Sure.

- [ ] Add @Pzixel to Cargo.toml.

README.md
vadixidav: Alright, I will see what information I can preserve/add in the `README`.

Cargo.toml
vadixidav: `system` tells `cv-sys` to use the installed OpenCV. By default, it is off. This means it builds OpenCV for you so you don't need to do anything. It also statically links OpenCV so you don't need any shared libraries/DLLs, just the binary/exe file.

.travis.yml
vadixidav: Cargo sets CMake to use the number of threads equal to whatever the system has, so this does nothing in the new build. OpenCV is being built by Cargo now, rather than externally.

.ci/install_cuda.sh
vadixidav: Installing CUDA is currently left up to the user (in Travis CI it is being installed using the system package manager). By default the `cuda` feature is off. We can include instructions on how to install CUDA in the `README.md` for Linux and Windows if desired.

vadixidav:
I think this branch has gotten too far out of sync with master, so I am just going to close it.

PR #122, opened 2019-03-25
Improve text feature

kkk669:
@Pzixel's #103 was great work for Homebrew users like me, but it seems to be insufficient to build cv-rs.

I added the following modifications.

1. Added `text` to `tesseract` feature's requirements because `text` feature is required by `tesseract` feature
    - If you run `cargo build --features "tesseract"`, then `text` feature is automatically enabled.
1. Move `text.h` and `text.cc` to `text` directory and modify `build.rs`, so that they will be compiled only when `text` feature is enabled
    - Without this modification of `build.rs`, `cargo build` always requires `opencv2/text/ocr.hpp`.

Also, I tested the following environment.

- macOS 10.14.3
- Homebrew 2.0.5
- OpenCV 4.0.1 (`brew install opencv`)
- build command: `CXXFLAGS=-I/usr/local/include/opencv4 cargo build`


Pzixel:
Cool!

